PR #1, opened 2022-04-28
Support adding CA Certificates

CRogers:
## Before this PR
There's no way to get a CA certificate installed into the downloaded JDK's `lib/security/cacerts` root truststore. For companies that use TLS interception internally this means any test that reaches out to the internet will fail.

Going from people having their own JDKs with this manually added to our JDKs they have no/little control over means this could be a real sticking point basically immediately.

## After this PR
==COMMIT_MSG==
Provide a way to add CA certs to all installed JDKs.
==COMMIT_MSG==


gradle-jdks/src/test/groovy/com/palantir/gradle/jdks/JdksPluginIntegrationSpec.groovy
```diff
@@ -61,6 +66,76 @@ class JdksPluginIntegrationSpec extends IntegrationSpec {
...
+                }
+            }
+        '''.stripIndent(true)
+
+        when:
+
+        def stdout = runTasksSuccessfully(
+                'printCaTruststoreAliases',
+                // TODO: avoid resolving from root configuration somehow, removed in Gradle 8
+                '--warning-mode=none')
```
carterkozak: Sounds like we may not need to suppress warnings here either?
```suggestion
                'printCaTruststoreAliases')
```

CRogers: 👍🏻 yep, will fix it up with the conflicts

gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdkManager.java
```diff
@@ -72,8 +76,14 @@ public Path jdk(JdkSpec jdkSpec) {
                 copy.into(temporaryJdkPath);
             });
 
+            Path javaHome = findJavaHome(temporaryJdkPath);
+
+            jdkSpec.caCerts().caCerts().forEach((name, caCertFile) -> {
+                addCaCert(javaHome, name, caCertFile);
+            });
```
carterkozak, line 83: Hmm, given two projects which use azul-17 x86_64, what happens if both projects specify completely different certificates? Do we need to include the certificates in our cache key? I'm not sure that we should mutate the jvm in a potentially leaky way such that tests pass on project B only if you've run project A first.

gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdksExtension.java
```diff
@@ -31,6 +32,8 @@ public abstract class JdksExtension {
 
     protected abstract MapProperty<JavaLanguageVersion, JdkExtension> getJdks();
 
+    public abstract MapProperty<String, File> getCaCerts();
```
carterkozak: Might be cleaner to take PEM formatted CA certs as strings, that way we don't need to think about file I/O. iirc they can be fed to keytool via stdin when no `-file` is specified.

CRogers: Ah, nice spot! I did this initially but figured keytool just always needs a file

gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdkManager.java
CRogers: ~~I made the CA certificates configurable at the root level/all jdks level, so all the jdks will have the same CA certs.~~ misread what you wrote

CRogers: Certificates are included in the cache key:

https://github.com/palantir/gradle-jdks/blob/8c108193147da88532fe03c6e4bf41971d1828d6/gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdkSpec.java#L35-L40

carterkozak: Ah fantastic, I missed that in the block above :-)

gradle-jdks/src/main/java/com/palantir/gradle/jdks/CaCerts.java
```diff
@@ -0,0 +1,60 @@
...
+import java.nio.file.Files;
+import java.util.Comparator;
+import java.util.Map;
+import java.util.SortedMap;
+import java.util.TreeMap;
+import org.immutables.value.Value;
+
+@Value.Immutable
+interface CaCerts {
+    SortedMap<String, File> caCerts();
```
carterkozak: Theoretically SortedMap is obsolete in favor of [NavigableMap](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/NavigableMap.html) (which TreeMap implements)
```suggestion
    NavigableMap<String, File> caCerts();
```

carterkozak (approved):
lgtm!

PR #2, opened 2022-04-28
Fix the build

CRogers:
## Before this PR
Build does not work yet.

## After this PR
Build works


build.gradle
```diff
@@ -27,7 +27,6 @@ allprojects {
     repositories {
         gradlePluginPortal()
```
carterkozak, line 28: Do we need to opt out of metadata here as well?
```suggestion
        gradlePluginPortal() { metadataSources { mavenPom(); ignoreGradleMetadataRedirection() } }
```

carterkozak:
oof, the classpath conflicts are coming from baseline itself due to some eclipse workaround thing :-(

build.gradle
CRogers: I think an excavator just enforces this??

CRogers: hmm, this did not do what I expected https://github.com/palantir/gradle-jdks/pull/3

carterkozak (approved):

PR #10, opened 2022-04-29
Default storage location to `~/.gradle/caches/gradle-jdks`

CRogers:
## Before this PR
We saved the JDKs to `build/jdks` mainly so that these were easier to investigate while testing.

## After this PR
We save JDKs in `~/.gradle/caches/gradle-jdks`.

This means extracted JDKs can be shared between directories. 
Using this directory because we already cache `~/.gradle/caches` in most of our circle builds.

In tests we override it back to `build/jdks` so we actually exercise the codepaths we expect/better isolation.

### DIFF
--- a/gradle-jdks/src/test/groovy/com/palantir/gradle/jdks/JdksPluginIntegrationSpec.groovy
+++ b/gradle-jdks/src/test/groovy/com/palantir/gradle/jdks/JdksPluginIntegrationSpec.groovy
@@ -19,12 +19,17 @@ class JdksPluginIntegrationSpec extends IntegrationSpec {
                     distribution = 'azul-zulu'
                     jdkVersion = '11.54.25-11.0.14.1'    
                 }
+            
+                jdkStorageLocation = layout.buildDirectory.dir('jdks')
             }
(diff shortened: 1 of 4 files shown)


gradle-jdks/src/test/groovy/com/palantir/gradle/jdks/JdksPluginIntegrationSpec.groovy
```diff
@@ -53,21 +58,16 @@ class JdksPluginIntegrationSpec extends IntegrationSpec {
...
             }
-        '''.stripIndent(true)
+        '''.stripIndent(true), subprojectDir
 
         when:
-
-        def stdout = runTasksSuccessfully(
-                'printJavaVersion',
-                // TODO: avoid resolving from root configuration somehow, removed in Gradle 8
-                '--warning-mode=none')
```
CRogers, line 69: I really thought I encountered this deprecation warning - but seemingly I can't get it to reappear, even using a subproject 🤷🏻

carterkozak (approved):

PR #19, opened 2022-05-05
Allow lazier configuration of jdks

CRogers:
## Before this PR
When configuring an internal Gradle plugin to configure this plugin I ran into a problem - I couldn't work out which JDKs I had without doing an expensive network task - basically what @carterkozak warning me about in the baseline code.

## After this PR
==COMMIT_MSG==
Allow lazy configuration of JDKs
==COMMIT_MSG==

Extra methods allow me to provide all the jdk configuration lazily when it is requested.


gradle-jdks/src/main/java/com/palantir/gradle/jdks/DelayedConfigurableMap.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.Supplier;
+import org.gradle.api.Action;
+import org.gradle.api.provider.Provider;
+import org.gradle.api.provider.ProviderFactory;
+
+final class DelayedConfigurableMap<K, V> {
```
CRogers: I tried to use some of the built in `NamedDomainObjectCollection`/`MapProperty` but they really did not seem to do exactly what I wanted - so I build my own thing here.

gradle-jdks/src/main/java/com/palantir/gradle/jdks/DelayedConfigurableMap.java
```diff
@@ -0,0 +1,85 @@
...
+    private final ProviderFactory providerFactory;
+    private final Supplier<V> valueFactory;
+    private final List<Provider<Map<K, Action<V>>>> configurations = new ArrayList<>();
```
carterkozak: `>>>>` looks a lot like a merge conflict!

gradle-jdks/src/main/java/com/palantir/gradle/jdks/DelayedConfigurableMap.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.Supplier;
+import org.gradle.api.Action;
+import org.gradle.api.provider.Provider;
+import org.gradle.api.provider.ProviderFactory;
+
+final class DelayedConfigurableMap<K, V> {
```
carterkozak: What would it look like to abandon the map concept entirely, in favor of something shaped more like a function? I suspect this isn't the only place the map property may bite us

gradle-jdks/src/main/java/com/palantir/gradle/jdks/DelayedConfigurableMap.java
CRogers: This is how you tell it's good code

gradle-jdks/src/main/java/com/palantir/gradle/jdks/DelayedConfigurableMap.java
CRogers: I agree we should probably do that, started it here: https://github.com/palantir/gradle-baseline/pull/2263

CRogers: I updated this PR to show what it would look like.

PR #40, opened 2022-05-17
Add x86_64 to arch list

CRogers:
## Before this PR
Missed out `x86_64` arch string for macos.

## After this PR
==COMMIT_MSG==
Prevent `Cannot get architecture for x86_64` errors on macos.
==COMMIT_MSG==


xRuiAlves (approved):
Thanks for fixing! 👍

PR #41, opened 2022-05-17
The Azul JDK distribution uses tar.gz everywhere except windows

carterkozak:
==COMMIT_MSG==
The Azul JDK distribution uses tar.gz everywhere except windows, matching available distributions from the Azul CDN
==COMMIT_MSG==

### DIFF
--- a/gradle-jdks/src/main/java/com/palantir/gradle/jdks/AzulZuluJdkDistribution.java
+++ b/gradle-jdks/src/main/java/com/palantir/gradle/jdks/AzulZuluJdkDistribution.java
@@ -37,7 +37,10 @@ final class AzulZuluJdkDistribution implements JdkDistribution {
                 os(jdkRelease.os()),
                 arch(jdkRelease.arch()));
 
-        return JdkPath.builder().filename(filename).extension(Extension.ZIP).build();
+        return JdkPath.builder()
+                .filename(filename)
+                .extension(extension(jdkRelease.os()))
+                .build();
     }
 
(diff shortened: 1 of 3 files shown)


gradle-jdks/src/main/java/com/palantir/gradle/jdks/AzulZuluJdkDistribution.java
```diff
@@ -66,6 +69,18 @@ private static String arch(Arch arch) {
...
+    private static Extension extension(Os operatingSystem) {
+        switch (operatingSystem) {
+            case MACOS:
+                // fall-through
+            case LINUX:
+                return Extension.TARGZ;
+            case WINDOWS:
+                return Extension.ZIP;
+        }
+        throw new IllegalArgumentException("Unknown OS: " + operatingSystem);
```
pkoenig10: nit: We use `UnsupportedOperationException` above but `IllegalArgumentException` here. I don't have a preference, but we should be consistent.

pkoenig10 (approved):

gradle-jdks/src/main/java/com/palantir/gradle/jdks/AzulZuluJdkDistribution.java
carterkozak:
```suggestion
        throw new UnsupportedOperationException("Unknown OS: " + operatingSystem);
```
Good call, thanks!

PR #42, opened 2022-05-19
Don't allow baseline-java-versions to be configured multiple times

mglazer:
## Before this PR

I work in a large monorepo, where occasionally indirectly including this plugin (transitively through another plugin) causes the BaselineJavaVersions plugin to be applied more than once. This isn't safe to do, so attempt to add a check that at least minimizes this potential edge case.

## After this PR
==COMMIT_MSG==
Don't allow baseline-java-versions to be configured multiple times
==COMMIT_MSG==

### DIFF
--- a/gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdksPlugin.java
+++ b/gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdksPlugin.java
@@ -38,7 +38,9 @@ public final class JdksPlugin implements Plugin<Project> {
         JdkManager jdkManager = new JdkManager(
                 jdksExtension.getJdkStorageLocation(), jdkDistributions, new JdkDownloaders(jdksExtension));
 
-        rootProject.getPluginManager().apply(BaselineJavaVersions.class);
+        if (!rootProject.getPluginManager().hasPlugin("com.palantir.baseline-java-versions")) {
+            rootProject.getPluginManager().apply(BaselineJavaVersions.class);
+        }
 
         rootProject
                 .getExtensions()


mglazer:
Is there a changelog bot for this repo?

gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdksPlugin.java
```diff
@@ -38,7 +38,9 @@ public void apply(Project rootProject) {
         JdkManager jdkManager = new JdkManager(
                 jdksExtension.getJdkStorageLocation(), jdkDistributions, new JdkDownloaders(jdksExtension));
 
-        rootProject.getPluginManager().apply(BaselineJavaVersions.class);
```
CRogers, line 41: Uh, isn't plugin application idempotent? This should be safe. Is your large monorepo doing some classloader shenanigans? Or using the `plugins {}` block widely?

CRogers: https://docs.gradle.org/current/dsl/org.gradle.api.plugins.PluginManager.html#org.gradle.api.plugins.PluginManager:apply(java.lang.Class)

> void apply(Class<?> type)
>
> Applies the given plugin. Does nothing if the plugin has already been applied.

CRogers: You could probably look at the gradle build scan to get some idea of there are multiple `BaselineJavaVersions` plugins being applied, which would imply some classloader issues. Even with this change, I would expect that it would still go wrong as there would be different versions of the extension class.

mglazer:
This was a build problem, fixing the build problem ended up addressing the issue.

PR #46, opened 2022-05-30
musl/alpine linux support

CRogers:
## Before this PR
Internally a tool uses alpine linux, as opposed to the usual ubuntu. So we need to support downloading musl jdks.

## After this PR
==COMMIT_MSG==
Use musl/alpine JDKs on linux distributions that use musl rather than glibc (like alpine linux).
==COMMIT_MSG==


fawind (approved):
Thanks!

CRogers:
Actually managing to test this in the docker container of the alpine linux service, getting `> Failed to run ldd - exited with exit code 1` - going to investigate further.

PR #51, opened 2022-06-05
Synchronize Directory and Map property to avoid multithreading issues

mglazer:
## Before this PR

There was a very infrequent race condition in a large internal project, which would cause a build failure.

## After this PR
==COMMIT_MSG==
Synchronize Directory and Map property to avoid multithreading issues
==COMMIT_MSG==


gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdksExtension.java
```diff
@@ -43,10 +41,29 @@ public JdksExtension() {
...
                 new LazilyConfiguredMapping<>(() -> getObjectFactory().newInstance(JdkDistributionExtension.class));
         this.jdks = new LazilyConfiguredMapping<>(() -> getObjectFactory().newInstance(JdkExtension.class));
+        // there is an extremely subtle race condition whereby a property can be mid finalization in one
+        // worker thread, and being read from another worker thread, and if the finalization happens just
+        // as the other thread is already reading, it can cause very infrequent build failure. While
+        // the build failure itself is extremely rare, on very large builds with larger worker pools, these
+        // errors occur enough times that it is significant
+        // tracing the source of this race condition is difficult, so to just mitigate the problem,
+        // these two properties have had their methods synchronized (which arguably gradle should probably
+        // do since it's known that Property access is racy)
```
CRogers, line 51: Wow, this sounds like a nightmare! Thanks for finding it.

gradle-jdks/src/main/java/com/palantir/gradle/jdks/SynchronizedDirectoryProperty.java
```diff
@@ -0,0 +1,194 @@
...
+import javax.annotation.Nullable;
+import org.gradle.api.Transformer;
+import org.gradle.api.file.Directory;
+import org.gradle.api.file.DirectoryProperty;
+import org.gradle.api.file.FileCollection;
+import org.gradle.api.file.FileTree;
+import org.gradle.api.file.RegularFile;
+import org.gradle.api.provider.Provider;
+
+final class SynchronizedDirectoryProperty implements DirectoryProperty {
```
CRogers: My worry about doing this is that if Gradle ever add a method to this interface, everything explodes. I think we can do something similar with `Proxy`s to avoid that case by making the type at runtime.

gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdksExtension.java
CRogers: Just for my own sanity: do you think the case of multiple worker threads causing this race is a "normal" gradle problem or huge monorepo doing special things problem?

gradle-jdks/src/main/java/com/palantir/gradle/jdks/SynchronizedDirectoryProperty.java
CRogers: I tried it and it doesn't work - but for the same reason this code doesn't work. Gradle casting interfaces to their own internal types strikes again:

> java.lang.ClassCastException: class com.palantir.gradle.jdks.SynchronizedDirectoryProperty cannot be cast to class org.gradle.api.internal.provider.PropertyInternal

gradle-jdks/src/main/java/com/palantir/gradle/jdks/SynchronizedInterface.java
```diff
@@ -0,0 +1,50 @@
...
+package com.palantir.gradle.jdks;
+
+import java.lang.reflect.Proxy;
+import java.util.Arrays;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+final class SynchronizedInterface {
```
CRogers, line 26: My biggest concern with synchronising everything is causing a deadlock somehow.

mglazer: Ya....most of these method implementations are fairly simple, there are just some which change state of the internal object but do it in a very unsafe way. This is a lot cleaner than the monstrosity I had.

gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdksExtension.java
mglazer: Both, but I think severely exacerbated by the sheer size of this project (just configuring sometimes takes a minute). I'm trying to reduce this, but because of the nature of how this build occurs in CI, basically this configuration steps happens sometimes 100+ times when this build is sharded out for a single commit. So this problem happens a lot more frequently than I think any other project could possibly hit it.

I also saw a random NPE in my testing, so there are other issues, but this is the most common one:

```
Caused by: java.lang.NullPointerException
        at org.gradle.internal.Try.ofFailable(Try.java:41)
        at org.gradle.internal.serialization.Cached$Deferred.result(Cached.java:53)
        at org.gradle.internal.serialization.Cached$Deferred.get(Cached.java:48)
        at org.gradle.internal.jvm.inspection.JvmInstallationMetadata$DefaultJvmInstallationMetadata.hasCapability(JvmInstallationMetadata.java:159)
        at org.gradle.jvm.toolchain.internal.ToolchainMatcher.lambda$implementationPredicate$1(ToolchainMatcher.java:44)
        at org.gradle.jvm.toolchain.internal.ToolchainMatcher.test(ToolchainMatcher.java:35)
        at org.gradle.jvm.toolchain.internal.ToolchainMatcher.test(ToolchainMatcher.java:25)
        at org.gradle.jvm.toolchain.internal.JavaToolchainQueryService.query(JavaToolchainQueryService.java:83)
        at org.gradle.jvm.toolchain.internal.JavaToolchainQueryService.lambda$findMatchingToolchain$0(JavaToolchainQueryService.java:62)
        at org.gradle.jvm.toolchain.internal.JavaToolchainQueryService.lambda$findMatchingToolchain$1(JavaToolchainQueryService.java:62)
        at org.gradle.api.internal.provider.DefaultProvider.calculateOwnValue(DefaultProvider.java:72)
        at org.gradle.api.internal.provider.AbstractMinimalProvider.calculateValue(AbstractMinimalProvider.java:103)
        at org.gradle.api.internal.provider.TransformBackedProvider.calculateOwnValue(TransformBackedProvider.java:64)
        at org.gradle.api.internal.provider.AbstractMinimalProvider.get(AbstractMinimalProvider.java:84)
        at com.palantir.baseline.plugins.javaversions.JavaToolchains.lambda$forVersion$1(JavaToolchains.java:44)
        at com.palantir.baseline.plugins.javaversions.JavaToolchains.lambda$forVersion$2(JavaToolchains.java:39)
        at org.gradle.api.internal.provider.DefaultProvider.calculateOwnValue(DefaultProvider.java:72)
        at org.gradle.api.internal.provider.AbstractMinimalProvider.get(AbstractMinimalProvider.java:84)
        at com.palantir.baseline.plugins.javaversions.JavaInstallationMetadataWrapper.getInstallationPath(JavaInstallationMetadataWrapper.java:62)
        at org.gradle.api.tasks.compile.JavaCompile.applyToolchain(JavaCompile.java:331)
        at org.gradle.api.tasks.compile.JavaCompile.createBaseSpec(JavaCompile.java:324)
        at org.gradle.api.tasks.compile.JavaCompile.createSpec(JavaCompile.java:292)
        at org.gradle.api.tasks.compile.JavaCompile.compile(JavaCompile.java:142)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at org.gradle.internal.reflect.JavaMethod.invoke(JavaMethod.java:104)
````

mglazer: There are a couple of gradle issues that are open/sorta fixed which comment on how this can happen on large projects:

https://github.com/gradle/gradle/issues/12990

https://github.com/gradle/gradle/issues/12811

gradle-jdks/src/main/java/com/palantir/gradle/jdks/SynchronizedInterface.java
CRogers: Yeah, stepping through what methods are called on a projects to these types it doesn't seem that wild.

CRogers:
:+1:

carterkozak (approved):
thanks!

gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdksExtension.java
CRogers: For posterity: smoking gun from gradle's end: 

https://github.com/gradle/gradle/blob/fd5b272251fef0bd6b92250c1c190334883013f8/subprojects/model-core/src/main/java/org/gradle/api/internal/provider/AbstractProperty.java#L519-L520

PR #68, opened 2022-06-16
Add better support for Windows

mglazer:
## Before this PR

`java` doesn't exist on windows, it's `java.exe`. So this plugin fails on Windows.

## After this PR

==COMMIT_MSG==
Fixed an issue preventing this plugin from working on Windows.
==COMMIT_MSG==

### DIFF
--- a/gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdkManager.java
+++ b/gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdkManager.java
@@ -109,7 +109,7 @@ public final class JdkManager {
             return files.filter(file -> Files.isRegularFile(file)
                             // macos JDKs have a `bin/java` symlink to `Contents/Home/bin/java`
                             && !Files.isSymbolicLink(file)
-                            && file.endsWith(Paths.get("bin/java")))
+                            && file.endsWith(Paths.get("bin/" + SystemTools.java())))
                     .findFirst()
                     // JAVA_HOME/bin/java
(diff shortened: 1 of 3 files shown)


gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdkManager.java
```diff
@@ -109,7 +109,7 @@ private Path findJavaHome(Path temporaryJdkPath) {
             return files.filter(file -> Files.isRegularFile(file)
                             // macos JDKs have a `bin/java` symlink to `Contents/Home/bin/java`
                             && !Files.isSymbolicLink(file)
-                            && file.endsWith(Paths.get("bin/java")))
+                            && file.endsWith(Paths.get("bin/" + SystemTools.java())))
```
carterkozak: I suspect we may need to take the environment-specific separator into account:
```suggestion
                            && file.endsWith(Paths.get("bin" + FileSystems.getDefault().getSeparator() + SystemTools.java())))
```

Alternatively we could use 
```java
SystemTools.java().equals(file.getFileName().toString())
    && "bin".equals(file.getParent() /* null check? */ .getFileName().toString());
````

carterkozak: Paths.get may handle normalization for us, I'm not sure off the top of my head -- safest/easiest may be `Paths.get("bin", SystemTools.java())`

carterkozak (approved):
thanks!

PR #71, opened 2022-06-21
Attempt to move unpacked JDKs into place more safely

carterkozak:
==COMMIT_MSG==
Attempt to move unpacked JDKs into place more safely with replace-existing and an attempted atomic move
==COMMIT_MSG==


fawind (approved):
Makes sense, thanks!

PR #74, opened 2022-06-27
Remove self from build to fix publishing

carterkozak:
==COMMIT_MSG==
Remove self from build to fix publishing
==COMMIT_MSG==


CRogers (approved):

CRogers:
excavator will add this back

PR #76, opened 2022-06-27
Gradle can handle FS operations via sync

carterkozak:
## Before this PR
We handled file movement operations in plugin code. Everything seemed to work, except ours in some niche cases.

## After this PR
Bluf: let's make the filesystem gradle's problem.
Note that we're still handling locking to prevent issues when multiple processes attempt to unpack a JDK concurrently.
==COMMIT_MSG==
Gradle can handle FS operations via sync
==COMMIT_MSG==

## Possible downsides?
Previously we were able to use a move operation, which doesn't necessarily duplicate bytes on disk. The sync is a more expensive operation, however JDKs are (relatively) small, and once they are cached, the instance is reused.

### DIFF
--- a/gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdkManager.java
+++ b/gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdkManager.java
@@ -26,17 +26,16 @@ import java.io.IOException;
 import java.nio.channels.FileChannel;
 import java.nio.channels.FileLock;
 import java.nio.charset.StandardCharsets;
-import java.nio.file.AtomicMoveNotSupportedException;
-import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.nio.file.StandardCopyOption;
 import java.nio.file.StandardOpenOption;
 import java.util.UUID;
 import java.util.concurrent.locks.Lock;
 import java.util.stream.Stream;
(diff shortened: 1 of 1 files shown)


gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdkManager.java
```diff
@@ -98,30 +97,14 @@ public Path jdk(Project project, JdkSpec jdkSpec) {
...
-                        StandardCopyOption.REPLACE_EXISTING,
-                        StandardCopyOption.ATOMIC_MOVE);
-            } catch (AtomicMoveNotSupportedException ignored) {
-                Files.move(temporaryJavaHome, permanentJavaHome, StandardCopyOption.REPLACE_EXISTING);
+    private static void moveJavaHome(Project project, Path temporaryJavaHome, Path permanentJavaHome) {
+        project.sync(new Action<CopySpec>() {
+            @Override
+            public void execute(CopySpec copySpec) {
+                copySpec.from(temporaryJavaHome);
+                copySpec.into(permanentJavaHome);
```
CRogers, line 105: Do we think Gradle is going to do anything sensible here?

carterkozak: It does not. While the task itself succeeds for me, it clobbers the java that another task may be using to attempt a compile! Closing this out -- I've confirmed that #77 fixes it.

PR #86, opened 2022-07-29
Support Zulu bundle flags

j-baker:
Some Zulu distributions come with extra flags. I care about EA builds,
but there are other flags, which currently fail the string.split test.

This PR attempts to handle these flags on the Zulu part of the version
string. The Java version string can also contain -ea, but I assume the
'rest' of the string is the Java version string.

## Before this PR
JDK version `19.0.21-ea-19.0.0-ea.6` fails.

## After this PR
JDK version `19.0.21-ea-19.0.0-ea.6` passes.

## Possible downsides?
This PR assumes that the rest of the string is the java version. So e.g. '19.0.21-ea-THIS!IS-NOT!A!VERSION!' would pass. This is similar behaviour to as present since the non-sensical openjdk version could just as easily be '19.0.21-ea-THIS!IS!NOT!A!VERSION' which would have passed prior to this PR.

### DIFF
--- a/gradle-jdks/src/main/java/com/palantir/gradle/jdks/AzulZuluJdkDistribution.java
+++ b/gradle-jdks/src/main/java/com/palantir/gradle/jdks/AzulZuluJdkDistribution.java
@@ -16,11 +16,18 @@
 
 package com.palantir.gradle.jdks;
 
+import com.google.common.collect.ImmutableSet;
 import com.palantir.gradle.jdks.JdkPath.Extension;
 import com.palantir.gradle.jdks.JdkRelease.Arch;
+import java.util.Arrays;
+import java.util.List;
 import org.immutables.value.Value;
 
 final class AzulZuluJdkDistribution implements JdkDistribution {
+    // See https://docs.azul.com/core/zulu-openjdk/versioning-and-naming for flag details.
+    private static final ImmutableSet<String> BUNDLE_FLAGS =
+            ImmutableSet.of("ea", "embvm", "cp1", "cp2", "cp3", "c2", "criu", "cr", "crac");
+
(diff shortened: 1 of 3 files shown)


gradle-jdks/src/main/java/com/palantir/gradle/jdks/AzulZuluJdkDistribution.java
```diff
@@ -16,11 +16,18 @@
...
 import com.palantir.gradle.jdks.JdkPath.Extension;
 import com.palantir.gradle.jdks.JdkRelease.Arch;
+import java.util.Arrays;
+import java.util.List;
 import org.immutables.value.Value;
 
 final class AzulZuluJdkDistribution implements JdkDistribution {
+    // See https://docs.azul.com/core/zulu-openjdk/versioning-and-naming for flag details.
+    private static final ImmutableSet<String> BUNDLE_FLAGS =
+            ImmutableSet.of("ea", "embvm", "cp1", "cp2", "cp3", "c2", "criu", "cr", "crac");
```
CRogers, line 29: There's a `criu` version of the JDK?? Interesting....

CRogers (approved):

gradle-jdks/src/main/java/com/palantir/gradle/jdks/AzulZuluJdkDistribution.java
j-baker: worth saying: some of these flags are discontinued! (embvm is one). One only works for 32-bit ARM. The docs are very clear that just because a flag exists doesn't necessarily mean you can acquire a JVM with it set!

PR #88, opened 2022-07-29
Omit the availability specifier if the user specified it

j-baker:
## Before this PR
If an -ea release is specified, the -ca specifier should not be.

## After this PR
If the user specifies an availability specifier (e.g. -ea, -sa, -ca) then the standard suffix of -ca is not appended.

## Possible downsides?

It's probably easier to choose a subscriber availability release.

### DIFF
--- a/gradle-jdks/src/test/groovy/com/palantir/gradle/jdks/AzulZuluJdkDistributionTest.java
+++ b/gradle-jdks/src/test/groovy/com/palantir/gradle/jdks/AzulZuluJdkDistributionTest.java
@@ -35,8 +35,8 @@ class AzulZuluJdkDistributionTest {
     @Test
     void zulu_version_combination_supports_bundle_flags() {
         ZuluVersionSplit versionSplit = AzulZuluJdkDistribution.splitCombinedVersion(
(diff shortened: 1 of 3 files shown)


gradle-jdks/src/test/groovy/com/palantir/gradle/jdks/AzulZuluJdkDistributionTest.java
```diff
@@ -53,6 +53,19 @@ void jdk_path_linux_x86_64() {
...
+    @Test
+    void jdk_path_linux_x86_64_early_access() {
+        AzulZuluJdkDistribution distribution = new AzulZuluJdkDistribution();
+        String version = ZuluVersionUtils.combineZuluVersions("19.0.21-ea", "19.0.0-ea.6");
+        JdkPath path = distribution.path(JdkRelease.builder()
+                .arch(Arch.X86_64)
+                .os(Os.LINUX_GLIBC)
+                .version(version)
+                .build());
+        assertThat(path.filename()).isEqualTo("zulu19.0.21-ea-jdk19.0.0-ea.6-linux_x64");
```
j-baker, line 65: copied directly from the internal azul mirror...

carterkozak (approved):
lgtm

PR #141, opened 2022-12-25
Improvement: Support GraalVM CE distributions

johnhany97:
## Before this PR
One could only use Azul or Amazon's JDKs. To use the GraalVM CE distribution, I had to disable the java versions tooling so that I can override `JAVA_HOME` that Gradle was using.

## After this PR

==COMMIT_MSG==
Support GraalVM CE distributions
==COMMIT_MSG==

## Possible downsides?
I'm unsure if musl/alpine linux is actually supported for GraalVM so I didn't implement that case. There seems to be no mention explicitly of it in https://www.graalvm.org/latest/docs/introduction/#available-distributions

### DIFF
--- a/changelog/@unreleased/pr-141.v2.yml
+++ b/changelog/@unreleased/pr-141.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Support GraalVM CE distributions
+  links:
+  - https://github.com/palantir/gradle-jdks/pull/141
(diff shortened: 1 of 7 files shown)


johnhany97:
For context, releases of GraalVM CE distributions can be seen here: https://github.com/graalvm/graalvm-ce-builds/releases/

I could add more validations around which combinations of platforms & java versions actually exist.

CRogers:
:+1:

PR #161, opened 2023-02-13
Produce Palantir CA Plugin

CRogers:
## Before this PR
We had a class that enabled other plugins to call it and it would include the Palantir corporate CA cert into JDKs, if it exists in the system truststore. This enables people to run tests in open source projects that hit public webservers from within the corporate VPN.

However, none of the other plugins (`gradle-jdks-latest` or `gradle-jdks-internal`) actually enable this.

## After this PR
We now produce a plugin rather than a class so multiple other plugins can apply it without worrying if it's already been applied.

==COMMIT_MSG==
Produce `com.palantir.jdks.palantir-ca` plugin for use by other Gradle plugins to enable using open source projects from within the corporate VPN.
==COMMIT_MSG==

### DIFF
--- a/gradle-jdks/src/main/java/com/palantir/gradle/jdks/PalantirCaPlugin.java
+++ b/gradle-jdks/src/main/java/com/palantir/gradle/jdks/PalantirCaPlugin.java
@@ -41,11 +41,18 @@ import org.gradle.process.ExecResult;
 public final class PalantirCaPlugin implements Plugin<Project> {
     private static final BigInteger PALANTIR_3RD_GEN_SERIAL = new BigInteger("18126334688741185161");
 
-    public void apply(Project rootProject) {
-        if (rootProject.getRootProject() != rootProject) {
+    private Project rootProject;
+    private PalantirCaExtension extension;
+
+    public void apply(Project possibleRootProject) {
+        if (possibleRootProject.getRootProject() != possibleRootProject) {
             throw new IllegalArgumentException(
                     "com.palantir.jdks.palantir-ca must be applied to the root project only");
         }
+
+        rootProject = possibleRootProject;
+
+        extension = rootProject.getExtensions().create("palantirCa", PalantirCaExtension.class);
 
         rootProject.getPluginManager().apply(JdksPlugin.class);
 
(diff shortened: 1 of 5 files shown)


gradle-jdks/src/main/java/com/palantir/gradle/jdks/PalantirCaPlugin.java
```diff
@@ -34,46 +34,54 @@
...
 import java.util.Optional;
+import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.process.ExecResult;
 
-public final class PalantirCa {
+public final class PalantirCaPlugin implements Plugin<Project> {
     private static final BigInteger PALANTIR_3RD_GEN_SERIAL = new BigInteger("18126334688741185161");
 
-    public static void applyToRootProject(Project rootProject, boolean strict) {
```
felixdesouza, line 43: should `strict` still not be configurable via some extension? the failures will effectively be silent as most builds don't run with `--info` ?

CRogers: We want the failures to be silent, at least on open source, as external contributors who do not have the palantir CA in their truststore need to run the code.

CRogers: Notable: public circleci

CRogers: Internally, it's a different matter - maybe it's good to have this explode if it can't find the cert. However, people shouldn't really be anything other than our internal mirror anyway. We'd probably need to have support for windows if we did do this.

CRogers: Maybe we should warn rather than info internally, so at least it's obvious...

CRogers: It now has configurable log levels, which we can set to warn internally, so people will bug us if this starts to fail at least rather than going unnoticed.

felixdesouza (approved):

PR #163, opened 2023-02-14
Don't fail when linux ca-bundle file does exist for Palantir CAs. Also support centos

CRogers:
## Before this PR
Internally, a service has a build that uses a redhat based linux rather than ubuntu. It does not have `/etc/ssl/certs/ca-certificates.crt`. Instead it has:

```
$ ls -l /etc/ssl/certs/
total 16
lrwxrwxrwx 1 root root   49 Sep 30  2021 ca-bundle.crt -> /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
lrwxrwxrwx 1 root root   55 Sep 30  2021 ca-bundle.trust.crt -> /etc/pki/ca-trust/extracted/openssl/ca-bundle.trust.crt
```

It fails with an error that:

```
Caused by: java.nio.file.NoSuchFileException: /etc/ssl/certs/ca-certificates.crt
	at com.palantir.gradle.jdks.PalantirCaPlugin.linuxSystemCertificates(PalantirCaPlugin.java:124)
```

## After this PR
==COMMIT_MSG==
* Support reading Palantir CA from the truststores of Redhat based systems
* Do not fail if the system truststore cannot be found on linux
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-163.v2.yml
+++ b/changelog/@unreleased/pr-163.v2.yml
@@ -1,0 +1,7 @@
+type: fix
+fix:
+  description: |-
+    * Support reading Palantir CA from the truststores of Redhat based systems
+    * Do not fail if the system truststore cannot be found on linux
+  links:
+  - https://github.com/palantir/gradle-jdks/pull/163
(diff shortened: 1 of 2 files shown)


felixdesouza:
Does Files.exists check whether the link exists or the underlying file?

felixdesouza:
👍 was on a phone when I reviewed, but it *does* follow links, LGTM

PR #168, opened 2023-02-20
Fix "Duplicate extensions not allowed" when reading the Palantir CA

CRogers:
Slack context: search slack for this PR link

## Before this PR
An internal macos user was getting errors like:

```
java.security.cert.CertificateParsingException: java.io.IOException: Duplicate extensions not allowed
        at com.palantir.gradle.jdks.PalantirCaPlugin.parseCerts(PalantirCaPlugin.java:155)
        ... 154 more
Caused by: java.io.IOException: Duplicate extensions not allowed
        ... 155 more
```

when using 0.30.0.

Upon inspecting their system truststore, there was an Apple KDC certificate in it which had duplicate extensions:

```
X509v3 extensions:
  X509v3 Key Usage:
      Digital Signature, Key Encipherment
  X509v3 Extended Key Usage:
      TLS Web Server Authentication
  X509v3 Extended Key Usage:
      1.3.6.1.5.2.3.5
```

This was causing Java to throw an exception (see https://bugs.openjdk.org/browse/JDK-8211712).

## After this PR
==COMMIT_MSG==
For the Palantir CA plugin, skip over certificates in the system truststore if they contain duplicate extensions, to avoid the `Duplicate extensions not allowed` error message.
==COMMIT_MSG==

## Possible downsides?
I've had to do some "fun" exception message matching to get this to work 🙃


gradle-jdks/src/main/java/com/palantir/gradle/jdks/PalantirCaPlugin.java
```diff
@@ -149,13 +150,36 @@ private static Optional<String> selectPalantirCertificate(byte[] multipleCertifi
                 .map(PalantirCaPlugin::encodeCertificate);
     }
 
-    private static Collection<? extends Certificate> parseCerts(byte[] multipleCertificateBytes) {
+    // visible for testing
```
felixdesouza: just use the annotation?

gradle-jdks/src/main/java/com/palantir/gradle/jdks/PalantirCaPlugin.java
```diff
@@ -149,13 +150,36 @@ private static Optional<String> selectPalantirCertificate(byte[] multipleCertifi
...
+
+        List<Certificate> certs = new ArrayList<>();
+
+        ByteArrayInputStream baos = new ByteArrayInputStream(multipleCertificateBytes);
+
+        for (int i = 0; baos.available() != 0; i++) {
+            try {
+                certs.add(certificateFactory.generateCertificate(baos));
+            } catch (CertificateException e) {
+                if (e.getMessage().contains("Duplicate extensions not allowed")) {
```
felixdesouza, line 171: funnn

gradle-jdks/src/main/java/com/palantir/gradle/jdks/PalantirCaPlugin.java
CRogers: I'm not going to bring in Guava for just that

CRogers: Oh wait, this uses guava. Sad :(

gradle-jdks/src/test/resources/amazon-cas-with-whitespace-between.pem
```diff
@@ -0,0 +1,57 @@
+https://www.amazontrust.com/repository/SFSRootCAG2.pem
```
felixdesouza, line 1: didn't realise you could put non cert text in a pem file

CRogers: I am a firm believer that more people should annotate what the certs actually are so you can quickly browse it

felixdesouza: yeah definitely, that's amazing

felixdesouza (approved):

PR #178, opened 2023-03-28
Handle incorrectly versioned certificates in system truststore

CRogers:
## Before this PR
Similar to #168. Someone internally has this x509 cert in their system truststore issued by StrongLoop. The cert uses v3 extensions but only lists itself as being v1. Java is [very strict on this](https://blog.thewatertower.org/2020/07/30/tolerant-and-intolerant-tls-clients-version-1-certificates-with-version-3-extensions/):

```
Caused by: java.security.cert.CertificateParsingException: no more data allowed for version 1 certificate
	at java.base/sun.security.x509.X509CertInfo.parse(X509CertInfo.java:679)
```

The printed form of the cert:

```
$ openssl x509 -in gradle-jdks/src/test/resources/strongloop-cert-with-v3-extensions-but-v1-version.pem -text
Certificate:
    Data:
        Version: 1 (0x0)
        Serial Number: 6606 (0x19ce)
    Signature Algorithm: sha1WithRSAEncryption
        Issuer: C=US, ST=CA, L=San Francisco, O=StrongLoop, Inc., OU=StrongOps, CN=ca.strongloop.com
        Validity
            Not Before: Dec  8 23:35:33 2015 GMT
            Not After : Apr 24 23:35:33 2043 GMT
        Subject: CN=strongloop.com
        Subject Public Key Info:
            Public Key Algorithm: rsaEncryption
                RSA Public-Key: (2048 bit)
                Modulus:
                    // snip
                Exponent: 65537 (0x10001)
        X509v3 extensions:
            X509v3 Subject Alternative Name:
                IP Address:0.0.0.0, IP Address:127.0.0.1
```

## After this PR
==COMMIT_MSG==
Fix errors of the form `no more data allowed for version 1 certificate` when the system truststore has incorrectly versioned X.509 certs in it.
==COMMIT_MSG==

We just ignore these ones - they are not going to be the Palantir CA.


felixdesouza (approved):

gradle-jdks/src/main/java/com/palantir/gradle/jdks/PalantirCaPlugin.java
```diff
@@ -173,6 +173,10 @@ static List<Certificate> parseCerts(byte[] multipleCertificateBytes) {
                     continue;
                 }
 
+                if (e.getMessage().contains("no more data allowed for version 1 certificate")) {
```
carterkozak, line 176: we may want to check that `e.getMessage()` doesn't return null prior to the `.contains` invocation

PR #184, opened 2023-04-02
Look for `javac` instead of `java`

### LINKED ISSUE #97: Support using the plugin with Java 8
## What happened?

I tried to add the plugin to [our project FitNesse](https://github.com/unclebob/fitnesse). I added the plugin to the `plugins` block and added the following config
```
jdks {
  jdk(8) {
    distribution = 'azul-zulu'
    jdkVersion = '8.64.0.19'
  }
}

javaVersions {
  libraryTarget = 8
}
```

When I tried to build I got: 
> java.lang.UnsupportedClassVersionError: com/palantir/gradle/jdks/JdksPlugin has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
> com/palantir/gradle/jdks/JdksPlugin has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0

## What did you want to happen?

I would have liked to be able to use this plugin also when my gradle is using Java 8

arlaneenalra:
There are some situations where using `bin/java` as a heuristic to find JAVA_HOME within a JDK package can result in mistakenly finding a JRE root instead. Looking for `bin/javac` fixes this since JREs do not have `javac`.

This should fix https://github.com/palantir/gradle-jdks/issues/97

## Before this PR

The `findJavaHome` method could mistakenly find a JRE inside of the JDK packages instead of the JDK resulting in an unusable JAVA_HOME. This main impacted some older JDKs.

## After this PR

==COMMIT_MSG==
Use `bin/javac` to find JAVA_HOME instead of `bin/java` to avoid accidentally installing a JRE instead of a JDK.
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-184.yml
+++ b/changelog/@unreleased/pr-184.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Use bin/javac to find JAVA_HOME instead of bin/java to avoid accidentally installing a JRE instead of a JDK.
+  links:
+  - https://github.com/palantir/gradle-jdks/pull/184
--- a/gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdkManager.java
+++ b/gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdkManager.java
@@ -194,7 +194,7 @@ public final class JdkManager {
             return files.filter(file -> Files.isRegularFile(file)
                             // macos JDKs have a `bin/java` symlink to `Contents/Home/bin/java`
                             && !Files.isSymbolicLink(file)
-                            && file.endsWith(Paths.get("bin", SystemTools.java())))
+                            && file.endsWith(Paths.get("bin", SystemTools.javac())))
                     .findFirst()
                     // JAVA_HOME/bin/java
                     .orElseThrow(() -> new RuntimeException("Failed to find java home in " + temporaryJdkPath))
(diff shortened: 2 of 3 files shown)


palantirtech:
Thanks for your interest in [palantir/gradle-jdks](https://github.com/palantir/gradle-jdks), @arlaneenalra! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

arlaneenalra:
Note: I'll sign the CLA .. but the rest of those items seem a bit much ..

arlaneenalra:
Bump: What is needed to get this reviewed?

jonahbeckford:
My error is different from the summary's linked issue #97, but I think this PR would fix the wrong selection of the JRE rather than the JDK when I'm running into with Java 8:

```
> Task :ffi-java:compileJava FAILED
Compiling with JDK Java compiler API.
FAILURE: Build failed with an exception.
* What went wrong:
Execution failed for task ':ffi-java:compileJava'.
> Could not find tools.jar. Please check that /root/.gradle/gradle-jdks/azul-zulu-8.72.0.17-8.0.382-569852f9b038db1a contains a valid JDK installation.
```

@arlaneenalra: It looks like you have to press two checkmark boxes ("Type: Fix" and "Check the box to generate changelog(s)"). Is that possible?

arlaneenalra:
> @arlaneenalra: It looks like you have to press two checkmark boxes ("Type: Fix" and "Check the box to generate changelog(s)"). Is that possible?

Maybe .. I have no idea how that changelog-app thing works and can't find the documentation in my cursory glance online .. I'll take a crack at adding what I think might be the correct file ...

PR #240, opened 2023-10-17
Gradle 8: Use dynamic proxy for `JavaInstallationMetadata`

CRogers:
## Before this PR
When used on Gradle 8, running `compileJava` or related java tasks leads to an error:

```
Execution failed for task ':<project>:compileJava'.
> Receiver class com.palantir.gradle.jdks.ImmutableGradleJdksJavaInstallationMetadata does not define or inherit an implementation of the resolved method 'abstract boolean isCurrentJvm()' of interface org.gradle.jvm.toolchain.JavaInstallationMetadata.
```

This is because Gradle have added a method to `JavaInstallationMetadata` with no default, probably not expecting people to implement it :(.

## After this PR
==COMMIT_MSG==
Support Gradle 8, avoid errors of the form "ImmutableGradleJdksJavaInstallationMetadata does not define or inherit an implementation of the resolved method 'abstract boolean isCurrentJvm()'"
==COMMIT_MSG==

Do this by making a dynamic proxy, which should work provided nothing actually calls these extra methods.

Related fix in baseline: https://github.com/palantir/gradle-baseline/pull/2605

## Possible downsides?
Wonderful hackery.


felixdesouza:
I'm not sure I understand, why do we have to reimplement this? Didn't we do this in https://github.com/palantir/gradle-baseline/pull/2605 ?

felixdesouza:
sorry, accidentally closed

CRogers:
We implement `JavaInstallationMetadata` in this repo too, so have the wrong incorrect inferface issue too.

felixdesouza:
👍

PR #286, opened 2024-02-16
Allow specifying different versions of JDKs per operating system and architecture

CRogers:
## Before this PR
Currently, gradle-jdks only allows you to specify one version of a JDK per java major version. This means all operating systems and architectures must use the same JDK version.

However, this falls apart in the presence of operating system only releases by JDK vendors. For example, Amazon Corretto recently released a "linux" only release for JDK 17 ([`17.0.10.8.1`](https://github.com/corretto/corretto-17/releases/tag/17.0.10.8.1)) and 21 ([`21.0.2.14.1`](https://github.com/corretto/corretto-21/releases/tag/21.0.2.14.1)).

We attempt to use same JDKs in both our prod environments and CI (which are linux based) and on developer's local machines (almost always macos). We had an unfortunate situation where automated upgrades of JDKs upgraded people's repos to these versions that did not have macos distributions, causing them to be unable to compile or test code locally.

## After this PR

==COMMIT_MSG==
Allow specifying different versions of JDKs per operating system and architecture
==COMMIT_MSG==

We still want to get the latest JDKs in prod, even if they're on a linux only release. Our compliance constraints mean we _must_ fix CVEs on a short timeframe, and these CVE fixes may be in a linux only release. Or there could be some critically important bugfix we need (but currently wouldn't be able to get without breaking people's dev setups). Even the current bug is scary and we should ensure gets onto prod ("AArch64: crypto pmull based CRC32/CRC32C intrinsics clobber V8-V15 registers").

There's changes internally to rest of the stack of Gradle plugins that deal with JDKs. Most notably, they all use the same json representation to list the latest JDKs, and we have an excavator that can produce this json file automatically.

This file format is in this repo because:
1. I want to make it easy to load into our Gradle configuration
2. When we build the feature to manage the Gradle daemon JDK, I want to produce a native-image binary that will "pre-install" the latest JDKs on CI images etc so the JDKs are already downloaded, which would necessitate the file format code living in this repo.

<details>
<summary>Example latest jdk json format</summary>

```json
{
  "jdksPerJavaVersion": {
    "8": {
      "distribution": "amazon-corretto",
      "os": {
        "macos": {
          "arch": {
            "x86-64": {
              "version": "8.402.08.1"
            },
            "aarch64": {
              "version": "8.402.08.1"
            }
          }
        },
        "linux-glibc": {
          "arch": {
            "x86-64": {
              "version": "8.402.08.1"
            },
            "aarch64": {
              "version": "8.402.08.1"
            }
          }
        },
        "linux-musl": {
          "arch": {
            "x86-64": {
              "version": "8.402.08.1"
            },
            "aarch64": {
              "version": "8.402.08.1"
            }
          }
        },
        "windows": {
          "arch": {
            "x86": {
              "version": "8.402.08.1"
            },
            "x86-64": {
              "version": "8.402.08.1"
            }
          }
        }
      }
    },
    "11": {
      "distribution": "amazon-corretto",
      "os": {
        "macos": {
          "arch": {
            "x86-64": {
              "version": "11.0.22.7.1"
            },
            "aarch64": {
              "version": "11.0.22.7.1"
            }
          }
        },
        "linux-glibc": {
          "arch": {
            "x86": {
              "version": "11.0.22.7.1"
            },
            "x86-64": {
              "version": "11.0.22.7.1"
            },
            "aarch64": {
              "version": "11.0.22.7.1"
            }
          }
        },
        "linux-musl": {
          "arch": {
            "x86-64": {
              "version": "11.0.22.7.1"
            },
            "aarch64": {
              "version": "11.0.22.7.1"
            }
          }
        },
        "windows": {
          "arch": {
            "x86": {
              "version": "11.0.22.7.1"
            },
            "x86-64": {
              "version": "11.0.22.7.1"
            }
          }
        }
      }
    },
    "17": {
      "distribution": "amazon-corretto",
      "os": {
        "macos": {
          "arch": {
            "x86-64": {
              "version": "17.0.10.7.1"
            },
            "aarch64": {
              "version": "17.0.10.7.1"
            }
          }
        },
        "linux-glibc": {
          "arch": {
            "x86-64": {
              "version": "17.0.10.8.1"
            },
            "aarch64": {
              "version": "17.0.10.8.1"
            }
          }
        },
        "linux-musl": {
          "arch": {
            "x86-64": {
              "version": "17.0.10.8.1"
            },
            "aarch64": {
              "version": "17.0.10.8.1"
            }
          }
        },
        "windows": {
          "arch": {
            "x86-64": {
              "version": "17.0.10.7.1"
            }
          }
        }
      }
    },
    "21": {
      "distribution": "amazon-corretto",
      "os": {
        "macos": {
          "arch": {
            "x86-64": {
              "version": "21.0.2.13.1"
            },
            "aarch64": {
              "version": "21.0.2.13.1"
            }
          }
        },
        "linux-glibc": {
          "arch": {
            "x86-64": {
              "version": "21.0.2.14.1"
            },
            "aarch64": {
              "version": "21.0.2.14.1"
            }
          }
        },
        "linux-musl": {
          "arch": {
            "x86-64": {
              "version": "21.0.2.14.1"
            },
            "aarch64": {
              "version": "21.0.2.14.1"
            }
          }
        },
        "windows": {
          "arch": {
            "x86-64": {
              "version": "21.0.2.13.1"
            }
          }
        }
      }
    }
  }
}
```

</details>


gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdkExtension.java
```diff
@@ -24,11 +30,46 @@ public abstract class JdkExtension {
 
     public abstract Property<JdkDistributionName> getDistributionName();
 
+    private final Map<Os, JdkOsExtension> jdkOsExtensions = new HashMap<>();
```
felixdesouza, line 33: why not use a `NamedDomainObjectContainer` ?

gradle-jdks/src/main/java/com/palantir/gradle/jdks/CurrentArch.java
```diff
@@ -0,0 +1,42 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.gradle.jdks;
+
+import java.util.Locale;
+import java.util.Set;
+
+final class CurrentArch {
+    static Arch current() {
```
felixdesouza: nit: since `Current*` is already in the name, how about `CurrentArch.get()`?

gradle-jdks/src/main/java/com/palantir/gradle/jdks/CurrentOs.java
```diff
@@ -27,15 +27,8 @@
...
 
-enum Os {
-    MACOS,
-    LINUX_GLIBC,
-    LINUX_MUSL,
-    WINDOWS;
-
-    @Value.Default
+final class CurrentOs {
     public static Os current() {
```
felixdesouza: same thing re `CurrentOs.get()`

gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdksPlugin.java
```diff
@@ -90,7 +90,15 @@ private JavaInstallationMetadata javaInstallationForLanguageVersion(
...
 
-        String version = jdkExtension.getJdkVersion().get();
+        Os currentOs = CurrentOs.current();
+        Arch currentArch = CurrentArch.current();
+
+        String version = jdkExtension
+                .jdkFor(currentOs)
+                .jdkFor(currentArch)
+                .getJdkVersion()
+                .get();
```
felixdesouza: curious, why force property retrieval here? do you not get into races with configuration?

gradle-jdks/src/test/groovy/com/palantir/gradle/jdks/JdksExtensionProjectSpec.groovy
```diff
@@ -0,0 +1,66 @@
...
+            println('jdkVersion linux-glibc x64: ' + jdkVersionFor(Os.LINUX_GLIBC, Arch.X86_64))
+        '''.stripIndent(true)
+
+        when:
+        def stdout = runTasksSuccessfully('help').standardOutput
+
+        then:
+        stdout.contains('jdkVersion macos aarch64: 11.1')
+        stdout.contains('jdkVersion linux-glibc aarch64: 11.2')
+        stdout.contains('jdkVersion linux-glibc x64: 11.3')
```
felixdesouza, line 64: nice

felixdesouza (commented):
mostly curious about mixing of non-provider/property types inside the extensions, but not a huge blocker

gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdkExtension.java
CRogers: With `NamedDomain....` need you to give a string for the name (or implement `Named`, which bring in a needless gradle dep). I don't really think it buys much? Unless there's some magic you can do with it I'm not aware of.

gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdksPlugin.java
CRogers: This method, `javaInstallationForLanguageVersion` is called from a lazy location, but you're right this could probably be even more lazy to avoid problems.

gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdkExtension.java
felixdesouza: Given how it's used, I think it doesn't buy us as much, since we need it to statically exist. I believe it creates most of the methods for you, but that doesn't really matter if you're calling it from java

gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdksPlugin.java
felixdesouza: ah

felixdesouza:
👍  pending `Current*.get()`

felixdesouza:
👍

Issue #333, opened 2024-03-21
Gradle JDK automanagement


crogoz:
## What happened?

gradle-jdks automanages the JDKs used for compilation, testing and running java programs from Gradle. It does not manage the JDK used to run Gradle.

## What did you want to happen?
gradle-jdks manages the JDK uses to run Gradle.

This means, when you run `./gradlew <someTask>` or open a Gradle project in IntelliJ, the correct JDK will be downloaded and used even if you have no JDK preinstalled. The correct certificates will also be set up.

PR #338, opened 2024-03-28
[Gradle JDK automanagment] Add gradlew patching logic

crogoz:
## After this PR

Part of: https://github.com/palantir/gradle-jdks/issues/333
- Patches the `./gradlew` script with the `./gradle-jdks-setup.sh` script. The patching is done by a task "wrapperJdkPatcher" that  finalized the "wrapper" task. 
- Adds an integrationTest that checks the full E2E workflow for running the patched `./gradlew` command
- Moves the integration tests subproject to the `gradle-jdks-setup/integrationTest` 
==COMMIT_MSG==
[Gradle JDK automanagment] Add gradlew patching logic
==COMMIT_MSG==


gradle-jdks/src/test/groovy/com/palantir/gradle/jdks/GradleJdkPatcherIntegrationTest.groovy
```diff
@@ -0,0 +1,201 @@
...
+    def setup() {
+
+        // language=groovy
+        buildFile << """
+            buildscript {
+                repositories {
+                    mavenCentral()
+                }
+                dependencies {
+                    classpath files(FILES)
```
CRogers, line 50: Why do we need to do this? I think this is (more-or-less) what nebula does?

gradle-jdks/src/test/groovy/com/palantir/gradle/jdks/GradleJdkPatcherIntegrationTest.groovy
```diff
@@ -0,0 +1,229 @@
...
+
+import java.nio.file.Files
+import java.nio.file.Path
+
+class GradleJdkPatcherIntegrationTest extends IntegrationSpec {
+
+    private static String GRADLE_7VERSION = "7.6.2";
+    private static String GRADLE_8VERSION = "8.4";
+    private static AmazonCorrettoJdkDistribution CORRETTO_JDK_DISTRIBUTION = new AmazonCorrettoJdkDistribution();
+    private static String CORRETTO_DISTRIBUTION_URL_ENV = "CORRETTO_DISTRIBUTION_URL";
```
CRogers: For these things that are just used in one place, I'd just inline them - not sure making them static adds much.

gradle-jdks/src/test/groovy/com/palantir/gradle/jdks/GradleJdkPatcherIntegrationTest.groovy
crogoz: Without this classpath addition I will be getting `plugin com.palantir.jdks` not found when running the `./gradlew command` (https://github.com/palantir/gradle-jdks/pull/338/files#diff-98ba650ac01c560cdf5ffae45e4a36b0f4786c3956183b2d9cc1d9759b3061ceR222)
When running `./gradlew` directly the classpath  is not injected.

gradle-jdks/src/test/groovy/com/palantir/gradle/jdks/GradleJdkPatcherIntegrationTest.groovy
CRogers: Perhaps the serial numbers make sense, as they're just a string of numbers and need a name.

gradle-jdks/src/test/groovy/com/palantir/gradle/jdks/GradleJdkPatcherIntegrationTest.groovy
CRogers: Ah, makes sense, as we not actually using nebula here.

gradle-jdks/src/main/java/com/palantir/gradle/jdks/GradleWrapperPatcher.java
```diff
@@ -0,0 +1,182 @@
...
+    private static void write(Path destPath, String content) {
+        try {
+            Files.write(destPath, content.getBytes(StandardCharsets.UTF_8));
+        } catch (IOException e) {
+            throw new RuntimeException("Unable to write file", e);
+        }
+    }
+
+    private static boolean isGradleJdkSetupEnabled(ProjectLayout projectLayout) {
+        FileCollection gradleProperties = projectLayout.files("gradle.properties");
```
CRogers: Rather than manually reading the `gradle.properties`, we should just use `Project.property(...)` to find it so we get [the hierarchy support](https://docs.gradle.org/8.7/userguide/build_environment.html).

The trick to do this in an autoparallelisable-friendly way is to add a param called `getGradleJdksEnabled()` and set it's default value to be something that looks up the property using the Project object.

CRogers: Actually, I wonder if we could just use `onlyIf` here instead

gradle-jdks/src/main/java/com/palantir/gradle/jdks/GradleWrapperPatcher.java
```diff
@@ -0,0 +1,182 @@
...
+
+    public abstract static class GradleWrapperPatcherTask extends GradleWrapperPatcherTaskImpl {}
+
+    static void action(Params params) {
+        if (!isGradleJdkSetupEnabled(params.getProjectLayout())) {
+            return;
+        }
+        log.lifecycle("Gradle JDK setup is enabled, patching the gradle wrapper files");
+        Path gradlewFile = params.getProjectLayout()
+                .files(GRADLEW_UNIX_SCRIPT)
```
CRogers: We should pass this in as a `@OutputFile RegularFileProperty getGradlewScript();`. I think it's not possible to make something both an input and output (worth trying though), so you may need to set `.upToDateWhen { false }` in the `GradleWrapperPatcherTask` constructor.

gradle-jdks/src/main/resources/gradlew-patch
CRogers, line 1: Would end this with `.sh` to get syntax highlighting.

CRogers (commented):
Looks good, just a few comments.

CRogers (approved):

PR #341, opened 2024-04-05
[Gradle JDK automanagment] GradleWrapperMain patching

crogoz:
## After this PR

Part of: #333 

Workflow changes: 

Modifies GradleWrapperMain when running `./gradlew wrapper`:
- extracts `gradle-wrapper.jar` content into a new directory
- renames the old class `GradleWrapperMain` into `OriginalGradleWrapperMain` using ASM
- Creates a new `GradleWrapperMain` class that delegates the installation of the jdks (see explanation below) to the `./gradle/gradle-jdks-setup.sh` script
- Disables the auto-provisioning of the JDKs ([gradle doc](https://docs.gradle.org/8.5/userguide/toolchains.html#sub:disable_auto_provision))& the autodetection([gradle docs](https://docs.gradle.org/8.5/userguide/toolchains.html#sub:disable_auto_detect))
- Sets the custom toolchain installation paths ([gradle docs](https://docs.gradle.org/8.5/userguide/toolchains.html#sec:custom_loc)) to the installed jdks. 
- creates the new `gradle-wrapper.jar` file

Modifies `./gradle/gradle-jdks-setup.sh` script:
- installs all the jdks that are configured in the `gradle/jdks` configuration directory
- appends a property(`running.from.gradlew`) to the `GradlwWrapperMain` args; this ensures that if GradleWrapperMain was already called through `./gradlew` we won't need to do any extra setup. 

Modifies the `./gradlew` patch to include the  `running.from.gradlew` property

Implementation changes: 
- new subproject `gradle-jdks-setup-common` that groups the classes (CurrentOs/CurrentArch/Os/Arch/CommandRunner etc.). It will be added in the fatJar `gradle-jdks-setup-all*.jar`

==COMMIT_MSG==
[Gradle JDK automanagment] GradleWrapperMain patching logic
==COMMIT_MSG==


gradle-jdks/src/main/java/com/palantir/gradle/jdks/OrigGradleWrapperCreator.java
```diff
@@ -0,0 +1,104 @@
...
+import org.objectweb.asm.ClassVisitor;
+import org.objectweb.asm.ClassWriter;
+import org.objectweb.asm.MethodVisitor;
+import org.objectweb.asm.Opcodes;
+import org.objectweb.asm.Type;
+
+public final class OrigGradleWrapperCreator {
+
+    private static final String OLD_CLASS_NAME = "org.gradle.wrapper.GradleWrapperMain";
+    private static final String NEW_CLASS_NAME = "org.gradle.wrapper.OrigGradleWrapper";
```
CRogers: Now we're using asm we can probably get away with calling this `OriginalGradleWrapperMain` rather than having to keep it the same string length.

gradle-jdks-setup/src/main/java/org/gradle/wrapper/GradleWrapperMain.java
```diff
@@ -0,0 +1,211 @@
...
+@SuppressWarnings("BanSystemOut")
+public final class GradleWrapperMain {
+
+    private GradleWrapperMain() {}
+
+    public static void main(String[] args) throws Exception {
+        configureJdkAutomanagement();
+
+        // Delegate back to the original GradleWrapperMain implementation
+        ClassLoader loader = new OriginalGradleWrapperLoader();
```
CRogers: Do we need to do this? `OrigGradleWrapper` is going to be in the same jar, I think we can just do `Class.forName("...")` and load it in the same classloader (I think Gradle does the classloader shenanigans in their GradleWrapperMain as it's loading a class from a different jar - the gradle distribution on disk).

gradle-jdks-setup/src/main/java/org/gradle/wrapper/GradleWrapperMain.java
```diff
@@ -0,0 +1,211 @@
...
+        Class<?> origGradleWrapperMainClass = loader.loadClass("org.gradle.wrapper.OrigGradleWrapper");
+        Method method = origGradleWrapperMainClass.getMethod("main", String[].class);
+        method.invoke(null, new Object[] {args});
+    }
+
+    private static void configureJdkAutomanagement() throws IOException {
+        // Delegate to gradle-jdks-setup.sh to install the JDK
+        // TODO(crogoz): maybe use java to do the setup
+        Path projectHome = projectHome();
+        CommandRunner.run(List.of("sh", "./gradle/gradle-jdks-setup.sh"), Optional.of(projectHome.toFile()));
```
CRogers: We should probably add a minor optimisation at some point to not do this if Gradle has been invoked from `./gradlew` - setting an environment variable or something we can check.

gradle-jdks-setup/src/main/java/org/gradle/wrapper/GradleWrapperMain.java
```diff
@@ -0,0 +1,211 @@
...
+        }
+    }
+
+    private static Path getGradleJdksPath() {
+        return Path.of(Optional.ofNullable(System.getenv("GRADLE_USER_HOME"))
+                        .orElseGet(() -> System.getProperty("user.home") + "/.gradle"))
+                .resolve("gradle-jdks");
+    }
+
+    public static String getArch() {
```
CRogers: These are the same as `CurrentOs` and `CurrentArch` etc right? Eventually we shouldn't need these in the Gradle plugin (only in the setup jar) after we delete the in-gradle downloading of JDKs, but until then they should probably live in some common project with `CommandRunner` to avoid duplicating the logic (then get fat jarred into here).

crogoz: Yes, I'll need to move these around to avoid duplication. I also need Os & Arch classes as well.

gradle-jdks/src/main/java/com/palantir/gradle/jdks/JarResources.java
```diff
@@ -0,0 +1,93 @@
...
+
+                if (jarEntry.isDirectory()) {
+                    outputFile.mkdirs();
+                } else {
+                    outputFile.getParentFile().mkdirs();
+
+                    try (FileOutputStream outputStream = new FileOutputStream(outputFile)) {
+                        byte[] buffer = new byte[2048];
+                        int bytesRead;
+                        while ((bytesRead = jarInputStream.read(buffer)) != -1) {
```
CRogers: Can just do `jarInputStream.transferTo(outputStream)`.

gradle-jdks/src/main/java/com/palantir/gradle/jdks/GradleWrapperPatcher.java
```diff
@@ -69,6 +90,59 @@ private static void patchGradlewContent(
...
+                addDirectoryToJar(jarOutputStream, file, newRelativePath);
+            } else {
+                String entryName = relativePath + file.getName();
+                JarEntry entry = new JarEntry(entryName);
+                jarOutputStream.putNextEntry(entry);
+
+                try (InputStream in = new FileInputStream(file)) {
+                    byte[] buffer = new byte[1024];
+                    int bytesRead;
+                    while ((bytesRead = in.read(buffer)) != -1) {
```
CRogers: Another `transferTo`.

gradle-jdks/src/main/java/com/palantir/gradle/jdks/GradleWrapperPatcher.java
```diff
@@ -69,6 +90,59 @@ private static void patchGradlewContent(
         write(patchedGradlewScript.getAsFile().get().toPath(), getNewGradlewWithPatchContent(linesNoPatch));
     }
 
+    private static void patchGradlewJar(
+            Path buildDir, RegularFileProperty originalGradleWrapperJar, RegularFileProperty patchedGradleWrapperJar) {
+        JarResources.extractJar(originalGradleWrapperJar.getAsFile().get(), buildDir);
+
+        OrigGradleWrapperCreator.create(buildDir);
+        String[] classPaths = System.getProperty("java.class.path").split(File.pathSeparator);
```
CRogers: Reading this from the classpath feels a bit funky as we have then have our `org.gradle.wrapper.GradleWrapperMain` in the gradle plugins classpath even though we don't use it. Shouldn't this already be in the repo somewhere? Can't remember how/where we create the `gradle/gradle-jdks-setup.jar`.  That or I think it would technically be better to either:

1) Include the gradle-wrapper-setup.jar in the resources of the this jar, so we can still read it easily (probably the easiest way).
2) Use Gradle to download the jar by making a `Configuration`, adding the dep with the correct version and resolving it.

gradle-jdks-setup/src/main/java/org/gradle/wrapper/GradleWrapperMain.java
```diff
@@ -0,0 +1,153 @@
...
+        // Delegate to gradle-jdks-setup.sh to install the JDK if it hasn't been installed yet
+        Path projectHome = projectHome();
+        if (!isRunningFromGradlew()) {
+            CommandRunner.run(List.of("./gradle/gradle-jdks-setup.sh"), Optional.of(projectHome.toFile()));
+        }
+
+        // Set the daemon Java Home
+        String osName = CurrentOs.get().uiName();
+        String archName = CurrentArch.get().uiName();
+        Path daemonKdkMajorVersionPath = projectHome.resolve("gradle/gradle-daemon-jdk-version");
```
CRogers: daemonJdk?

gradle-jdks-setup/src/main/java/org/gradle/wrapper/GradleWrapperMain.java
```diff
@@ -0,0 +1,153 @@
...
+            File gradleJdkConfigurationDir, Path gradleJdksInstallationDir, String localPathPattern)
+            throws IOException {
+        List<String> paths = new ArrayList<>();
+        File[] files = gradleJdkConfigurationDir.listFiles();
+        if (files == null) {
+            return paths;
+        }
+        for (File file : files) {
+            if (file.isDirectory()) {
+                paths.addAll(getAllInstallationsPaths(file, gradleJdksInstallationDir, localPathPattern));
```
CRogers: It might be better to use `Files.walk` or `Files.visitFileTree` here rather than manually doing the recursive file exploration.

CRogers: In fact, why is this searching the file tree - shouldn't the installation path be at a fixed known location inside `$projectRoot/gradle/jdks/....`? Why do the file searching?

gradle-jdks-setup/src/main/java/org/gradle/wrapper/GradleWrapperMain.java
```diff
@@ -0,0 +1,153 @@
...
+
+    private static Path wrapperJar() {
+        URI location;
+        try {
+            location = GradleWrapperMain.class
+                    .getProtectionDomain()
+                    .getCodeSource()
+                    .getLocation()
+                    .toURI();
+        } catch (URISyntaxException var3) {
```
CRogers: Probably give this a better name than `var3`, even if it's `e` (was this copied from a decompiler?)

gradle-jdks-setup/src/main/java/org/gradle/wrapper/GradleWrapperMain.java
```diff
@@ -0,0 +1,153 @@
...
+            throw new RuntimeException(var3);
+        }
+
+        if (!location.getScheme().equals("file")) {
+            throw new RuntimeException(
+                    String.format("Cannot determine classpath for wrapper Jar from codebase '%s'.", location));
+        } else {
+            try {
+                return Paths.get(location);
+            } catch (NoClassDefFoundError var2) {
```
CRogers: ditto

gradle-jdks/build.gradle
```diff
@@ -14,13 +14,16 @@ dependencies {
     api project(':gradle-jdks-distributions')
     api project(':gradle-jdks-json')
     api project(':gradle-jdks-setup')
+    api project(':gradle-jdks-setup-common')
```
CRogers: I think because the files that were moved to `gradle-jdks-setup-common` were originally in `gradle-jdks-distributions`, we should make it an api dep from `gradle-jdks-distributions` not here?

gradle-jdks/build.gradle
```diff
@@ -66,4 +69,5 @@ tasks.withType(JavaCompile).configureEach {
 
 tasks.test {
     environment("PROJECT_VERSION", project.version)
+    dependsOn ':gradle-jdks-setup:fatJar'
```
CRogers, line 71: `GradleJdkInstallationSetupIntegrationTest` is inside the `integrationTest` source set, so I think this dependsOn needs to go on under `tasks.named('integrationTest').configure {`

CRogers: Hmm, that's actually in `gradle-jdks-setup`

gradle-jdks-setup/build.gradle
```diff
@@ -21,10 +22,23 @@ dependencies {
     integrationTestImplementation 'org.assertj:assertj-core'
 }
 
+task fatJar(type: Jar) {
```
CRogers: Might as well use `tasks.register('fatJar', Jar) {` for Gradle best practices.

gradle-jdks-setup/build.gradle
```diff
@@ -21,10 +22,23 @@ dependencies {
...
+
 tasks.withType(JavaCompile) {
     options.errorprone.disable 'PreferSafeLoggableExceptions'
 }
 
 tasks.integrationTest {
     environment("PROJECT_VERSION", project.version)
 }
+
+tasks.build.dependsOn tasks.fatJar
```
CRogers, line 45: How are we going to ship the fat jar to people? I guess we don't have the code in the plugin to do this yet (?), but I imagine we're either going to include the fat jar as a resource or download it?

crogoz: Correct, we don't have yet the code in the plugin to do this yet. 
I think it needs to be distributed alongside "./gradle-jdks.setup.sh" and the "gradle/" files. I was thinking of downloading it (similar how the wrapper.jar gets added to a project). I'll do this in the next PR.

CRogers (approved):

PR #376, opened 2024-07-03
Bump intellij dependency

crogoz:
## Before this PR

Screenshot 2024-07-03 at 16 10 28

onStart method marked for removal.

## After this PR
Removing the deprecated `onStart` override & Bumping the intellij dependency
==COMMIT_MSG==
Bump intellij dependency
==COMMIT_MSG==

### DIFF
--- a/idea-plugin/src/main/java/com/palantir/gradle/jdks/GradleJdksExternalSystemTaskNotificationListener.java
+++ b/idea-plugin/src/main/java/com/palantir/gradle/jdks/GradleJdksExternalSystemTaskNotificationListener.java
@@ -35,9 +35,6 @@ public final class GradleJdksExternalSystemTaskNotificationListener implements E
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


idea-plugin/src/main/java/com/palantir/gradle/jdks/GradleJdksExternalSystemTaskNotificationListener.java
```diff
@@ -34,9 +34,6 @@ public void onStart(@NotNull ExternalSystemTaskId id, String _workingDir) {
         }
     }
 
-    @Override
-    public void onStart(@NotNull ExternalSystemTaskId _id) {}
-
     @Override
     public void onEnvironmentPrepared(@NotNull ExternalSystemTaskId _id) {}
```
CRogers: Can we remove this one too as it has a default impl (from before `onStart` had a default impl)?

CRogers:
:+1:

PR #381, opened 2024-07-15
[Gradle JDK Automanagement] Add install-jdks script and gradle jdk configuration generation 

crogoz:
## After this PR

Part of the Gradle JDK automanagement. 
We need to preinstall the latest versions of all used JDKs in the non-developer environments docker images to get cache hits.
- Added a class `GradleJdksConfigurator ` that can configure the JDK configuration files for some specific JDK version specified [by a set of parameters](https://github.com/palantir/gradle-jdks/pull/381/files#diff-7c1ba37b2bd964f62ad9de29190e8259bb4ee0784be548fbbf488320c8dcaa1bR45-R53).The script `install-jdks.sh` can install & configure the JDKs configured in the configuration files.

- Moved the common functions for the scripts: `install-jdks.sh` & `gradle-jdks-setup.sh` into a `gradle-jdks-functions.sh`

- the `install-jdks.sh` script re-uses the `gradle-jdks-setup.jar` when installing & configuring jdks

==COMMIT_MSG==
[Gradle JDK Automanagement] Add install-jdks.sh script and gradle jdk configuration generation 
==COMMIT_MSG==


crogoz:
Adding "DO NOT MERGE" until I test the `Gradle JDK Automanagement` excavator

gradle-jdks-excavator-configurations/src/main/resources/install-jdks.sh
```diff
@@ -0,0 +1,91 @@
...
+  read -r value < "$1" || die "ERROR: Unable to read value from $1. Make sure the file ends with a newline."
+  echo "$value"
+}
+
+GRADLE_JDKS_DIR=$1
+CERTS_DIR=$2
+
+if [[ -z "$CERTS_DIR" || -z "$GRADLE_JDKS_DIR" ]]; then
+  die "ERROR: Missing argument CERTS_DIR"
+fi
```
CRogers: This will never happen as due to `-u` the `CERTS_DIR=$1` will explode (and `-e` will cause the script to exit immediately).

gradle-jdks-excavator-configurations/src/main/resources/install-jdks.sh
```diff
@@ -0,0 +1,91 @@
...
+  Darwin* )         os_name="macos"  ;;       #(
+  * )               die "ERROR Unsupported OS: $( uname )" ;;
+esac
+
+if [ "$os_name" = "linux" ]; then
+    ldd_output=$(ldd --version 2>&1 || true)
+    if echo "$ldd_output" | grep -qi glibc; then
+       os_name="linux-glibc"
+    elif echo "$ldd_output" | grep -qi "gnu libc"; then
+           os_name="linux-glibc"
```
CRogers: odd indentation?

gradle-jdks-excavator-configurations/src/main/resources/install-jdks.sh
```diff
@@ -0,0 +1,91 @@
...
+    elif echo "$ldd_output" | grep -qi musl; then
+      os_name="linux-musl"
+    else
+      die "Unable to determine glibc or musl based Linux distribution: ldd_output: $ldd_output"
+    fi
+fi
+
+# Arch specific support, see: gradle-jdks:com.palantir.gradle.jdks.setup.common.CurrentArch.java
+case "$(uname -m)" in                         #(
```
CRogers: It feels to me like a lot of this logic should be shared with `gradle-jdks-setup.sh`? I'm worried that copying so much of the logic like this would mean they get out of sync.

gradle-jdks-excavator-configurations/src/main/java/com/palantir/gradle/jdks/GradleJdksConfigurator.java
```diff
@@ -0,0 +1,100 @@
...
+     * @param targetDir the target directory where the installation scripts should be written to.
+     */
+    public static void writeInstallationScripts(Path targetDir) {
+        Path scriptsDir = targetDir.resolve("scripts");
+        GradleJdksConfigsUtils.createDirectories(scriptsDir);
+        Path functionsScript = GradleJdksConfigsUtils.copyResourceToPath(scriptsDir, "gradle-jdks-functions.sh");
+        GradleJdksConfigsUtils.setExecuteFilePermissions(functionsScript);
+        Path installationScript = GradleJdksConfigsUtils.copyResourceToPath(scriptsDir, "install-jdks.sh");
+        GradleJdksConfigsUtils.setExecuteFilePermissions(installationScript);
+        GradleJdksConfigsUtils.copyResourceToPath(scriptsDir, "gradle-jdks-setup.jar");
```
CRogers, line 110: This is interesting - I assumed we were going to excavate out a version number to each repo and have them curl some tar, extract it then run the script. I think your change may be nicer though as we just extract everything the repo needs into the repo then they run it.

I think the interesting part is how we ensure we always update these files to latest. We can add this as a library to excavator but we need to make sure it keeps up to date. Otherwise we could download the latest version of this jar and load it in a classloader or something...

CRogers: Actually, I'm being silly excavator doesn't need releases and elevator will ensure it upgrades fine. Using as a library will work just fine.

gradle-jdks-excavator-configurations/src/main/java/com/palantir/gradle/jdks/GradleJdksConfigurator.java
```diff
@@ -0,0 +1,100 @@
...
+     * @param baseUrl the base url where the jdk distributions are hosted e.g. `https://corretto.aws`
+     * @param jdkDistributionName the name of the jdk distribution e.g. corretto.
+     * @param os the operating system of the jdk distribution e.g. linux.
+     * @param arch the architecture of the jdk distribution e.g. x64.
+     * @param javaVersion the major java version e.g. 8.
+     * @param jdkVersion the full version of the jdkVersion e.g. 8.282.08.1.
+     * @param caCerts the list of certificates that need to be added to the jdk. This influences the hash and the
+     * local installation directory name.
+     */
+    public static void writeGradleJdkInstallationConfigs(
```
CRogers: What do we expect to call this? If excavator how will it get all the correct data?

CRogers: It might be best if this takes the JSON representation that's somewhere else in this repo.

CRogers: [This one](https://github.com/palantir/gradle-jdks/blob/b59b494f8e29a302d70e518e2f3df4f3e9fa6bce/gradle-jdks-json/src/main/java/com/palantir/gradle/jdks/json/JdksInfoJson.java#L27)

gradle-jdks-excavator-configurations/src/main/resources/install-jdks.sh
```diff
@@ -0,0 +1,36 @@
...
+
+GRADLE_DIR=$1
+CERTS_DIR=$2
+
+# used by the resolved_symlink below to resolve the path based on the JAVA_VERSION value. e.g. /usr/local/${JAVA_VERSION}
+# shellcheck disable=SC2034
+SYMLINK_PATTERN=$3
+
+# Loading gradle jdk functions
+scripts_dir="$(dirname "$(readlink -f "$0")")"
```
CRogers: we normally use `$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )`, not sure why.

CRogers: I think it's best to use all caps for global variables and lower case for `local` variables, just to not mix them up

gradle-jdks-setup/src/main/resources/gradle-jdks-functions.sh
```diff
@@ -0,0 +1,156 @@
...
+    die "ERROR: $1 not found, aborting Gradle JDK setup"
+  fi
+  read -r value < "$1" || die "ERROR: Unable to read value from $1. Make sure the file ends with a newline."
+  echo "$value"
+}
+
+get_os() {
+  # OS specific support; same as gradle-jdks:com.palantir.gradle.jdks.setup.common.CurrentOs.java
+  case "$( uname )" in                          #(
+    Linux* )          os_name="linux"  ;;       #(
```
CRogers, line 51: This actually makes a global variable called `os_name`. If any other functions use os_name as variable it would get clobbered. Probably fine as bash is all single threaded, but best to declare this at the top as `local os_name` so it's scoped only to this function.

CRogers: Same for all the other function scoped variables here.

gradle-jdks-excavator-configurations/src/main/java/com/palantir/gradle/jdks/GradleJdksConfigurator.java
crogoz: > It might be best if this takes the JSON representation that's somewhere else in this repo.

Ah nice, I'll use this then. Didn't know the json representation was in gradle-jdks. I assumed the json representation was in excavator.

> What do we expect to call this? If excavator how will it get all the correct data?

I've done the parsing of the JSON representation in excavator and from there I was calling this method to just write the files.

gradle-jdks-setup/src/main/resources/gradle-jdks-functions.sh
crogoz: > but best to declare this at the top as local os_name so it's scoped only to this function.

It doesn't seem like local is supported in POSIX shells (these functions will be used by the `gradle-jdks-setup.sh` script as well which I think should be POSIX compliant). Should I change the names to `_<var_name>` instead: https://github.com/koalaman/shellcheck/wiki/SC3043 ?

CRogers: Ah, oh well. I guess `local` is bash-ism. I feel even with the underscore at the front these might clash between functions. We have tests to make sure we haven't messed it up (and there are no external callers of these functions), so it's probably fine.

CRogers (approved):

PR #382, opened 2024-07-17
[Gradle JDK] Fix when `palantir-gradle-jdks` is not installed and jdks are missing

crogoz:
## Before this PR

If a user didn't have the Intellij plugin `palantir-gradle-jdks` installed and if one of the jdks is not installed, they would get a failure in Intellij saying [Failed to find the toolchain at path= ...](https://github.com/palantir/gradle-jdks/compare/cr/fix-settings?expand=1#diff-1eebe5a89ca91bcce8ee9203a78fef241ebe4f6b593f82fa8b5e45fee28dbbe9L170). It isn't clear how to unblock.

## After this PR

- If one of the configured jdks is not installed, the settings plugin will call the `gradle/gradle-jdks-setup.sh` script. Thsi should only happen if the developer doesn't have the Intellij plugin installed.
- [Small cleanup, happy to move it to another PR] Changed the default log message for the Intellij Gradle JDK setup configuration run. Default was: [Process finished with exit code](https://github.com/xunfeng1980/intellij-community/blob/1812072b842bb808d84eb1b71fbd1ab6bd84cdf6/platform/ide-core/src/com/intellij/execution/process/ProcessTerminatedListener.java#L41C45-L41C81)
Screenshot 2024-07-17 at 6 29 14 PM

==COMMIT_MSG==
[Gradle JDK] Fix when `palantir-gradle-jdks` is not installed and jdks are missing
==COMMIT_MSG==

### DIFF
--- a/gradle-jdks-settings/src/main/java/com/palantir/gradle/jdks/settings/ToolchainJdksSettingsPlugin.java
+++ b/gradle-jdks-settings/src/main/java/com/palantir/gradle/jdks/settings/ToolchainJdksSettingsPlugin.java
@@ -25,13 +25,16 @@ import com.palantir.gradle.jdks.setup.common.CurrentArch;
 import com.palantir.gradle.jdks.setup.common.CurrentOs;
 import com.palantir.gradle.jdks.setup.common.Os;
 import java.io.IOException;
+import java.io.InputStream;
 import java.lang.reflect.Field;
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
+import java.nio.channels.FileChannel;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
@@ -73,10 +76,12 @@ public final class ToolchainJdksSettingsPlugin implements Plugin<Settings> {
         // Not failing here because the plugin might be applied before the `./gradlew setupJdks` is run, hence not
         // having the expected directory structure.
         if (!Files.exists(gradleJdksLocalDirectory)) {
-            logger.debug("Not setting the Gradle JDK properties because gradle/jdks directory doesn't exist. Please run"
-                    + " ./gradlew setupJdks to set up the JDKs.");
+            logger.lifecycle(
+                    "Not setting the Gradle JDK properties because gradle/jdks directory doesn't exist. Please run"
+                            + " ./gradlew setupJdks to set up the JDKs.");
             return;
         }
+        getInstalledJdkPaths(rootProjectDir, gradleJdksLocalDirectory);
         ProviderFactory providerFactory =
                 ((DefaultSettings) settings).getServices().get(ProviderFactory.class);
         if (!(providerFactory instanceof DefaultProviderFactory)) {
@@ -121,7 +126,7 @@ public final class ToolchainJdksSettingsPlugin implements Plugin<Settings> {
 
         @Override
         public Object invoke(Object _proxy, Method method, Object[] args) throws Throwable {
-            List<Path> installedLocalToolchains = getInstalledJdkPaths();
+            List<Path> installedLocalToolchains = getInstalledJdkPaths(rootProjectDir, gradleJdksLocalDirectory);
             // see: https://github.com/gradle/gradle/blob/4bd1b3d3fc3f31db5a26eecb416a165b8cc36082/subprojects/core-api/
             // src/main/java/org/gradle/api/internal/properties/GradleProperties.java#L28
             if (method.getName().equals("find") && args.length == 1) {
@@ -145,27 +150,22 @@ public final class ToolchainJdksSettingsPlugin implements Plugin<Settings> {
                 throw e.getCause();
             }
         }
+    }
 
-        private List<Path> getInstalledJdkPaths() {
-            List<Path> jdkPaths = getConfiguredJdkPaths(gradleJdksLocalDirectory);
-            List<Path> nonExistingJdkPaths =
-                    jdkPaths.stream().filter(path -> !Files.exists(path)).collect(Collectors.toList());
-            if (!nonExistingJdkPaths.isEmpty()) {
-                logger.error("Gradle JDK setup is enabled (palantir.jdk.setup.enabled is true) but some jdks were not"
-                        + " installed. If running from Intellij, please make sure the `palantir-gradle-jdks`"
-                        + " Intellij plugin is installed"
-                        + " https://plugins.jetbrains.com/plugin/24776-palantir-gradle-jdks/versions.");
-                runGradleJdkSetup(rootProjectDir);
-                if (jdkPaths.stream().anyMatch(path -> !Files.exists(path))) {
-                    throw new RuntimeException(String.format(
-                            "Some JDKs were not installed after running the setup script %s",
-                            jdkPaths.stream()
-                                    .filter(path -> !Files.exists(path))
-                                    .collect(Collectors.toList())));
-                }
-            }
-            return jdkPaths;
+    private static List<Path> getInstalledJdkPaths(Path rootProjectDir, Path gradleJdksLocalDirectory) {
+        List<Path> jdkPaths = getConfiguredJdkPaths(gradleJdksLocalDirectory);
+        List<Path> nonExistingJdkPaths =
+                jdkPaths.stream().filter(path -> !Files.exists(path)).collect(Collectors.toList());
+        if (!nonExistingJdkPaths.isEmpty()) {
+            logger.error(
+                    "Gradle JDK setup is enabled (palantir.jdk.setup.enabled is true) but some jdks were not"
+                            + " installed: {}. If running from Intellij, please make sure the"
+                            + " `palantir-gradle-jdks` Intellij plugin is installed"
+                            + " https://plugins.jetbrains.com/plugin/24776-palantir-gradle-jdks/versions.",
+                    nonExistingJdkPaths);
+            runGradleJdkSetup(rootProjectDir, nonExistingJdkPaths);
         }
+        return jdkPaths;
     }
 
     private static List<Path> getConfiguredJdkPaths(Path gradleJdksLocalDirectory) {
@@ -193,9 +193,13 @@ public final class ToolchainJdksSettingsPlugin implements Plugin<Settings> {
     }
 
     private static Path getToolchainInstallationDir() {
-        return Path.of(Optional.ofNullable(System.getenv("GRADLE_USER_HOME"))
-                        .orElseGet(() -> System.getProperty("user.home") + "/.gradle"))
-                .resolve("gradle-jdks");
+        return Path.of(getGradleUserHome()).resolve("gradle-jdks");
+    }
+
+    private static String getGradleUserHome() {
+        return Optional.ofNullable(System.getenv("GRADLE_USER_HOME"))
+                .orElseGet(() -> Optional.ofNullable(System.getenv("_TEST_GRADLE_USER_HOME"))
+                        .orElseGet(() -> System.getProperty("user.home") + "/.gradle"));
     }
 
     private static boolean isGradleVersionSupported() {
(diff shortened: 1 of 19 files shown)


gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/GradleJdkInstallationSetup.java
```diff
@@ -92,9 +92,7 @@ private static void setupDaemon(String[] args) {
         try {
             Files.createDirectories(projectDir.resolve(".gradle"));
             Path gradleConfigFile = projectDir.resolve(".gradle/config.properties");
-            if (!Files.exists(gradleConfigFile)) {
-                Files.createFile(gradleConfigFile);
-            }
+            gradleConfigFile.toFile().createNewFile();
```
crogoz, line 95: atomic creation

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/GradleJdkInstallationSetup.java
```diff
@@ -111,12 +109,14 @@ private static void setupJdk(StdLogger logger, CaResources caResources, String[]
         }
         Path destinationJdkInstallationDir = Path.of(args[1]);
         Path certsDir = Path.of(args[2]);
-        copy(logger, destinationJdkInstallationDir);
-        Map<String, String> certSerialNumbersToNames = extractCertsSerialNumbers(logger, certsDir);
-        caResources.maybeImportCertsInJdk(destinationJdkInstallationDir, certSerialNumbersToNames);
+        boolean wasCopied = copy(logger, destinationJdkInstallationDir);
+        if (wasCopied) {
+            Map<String, String> certSerialNumbersToNames = extractCertsSerialNumbers(logger, certsDir);
```
crogoz, line 117: if we didn't copy the jdk to the destination directory (https://github.com/palantir/gradle-jdks/pull/382/files#diff-d2a5b341108734345bf72bfb7dcb34cdc12976f8ffcf2b99df905702ce8faa12R130), which means that we waited for the lock while another process created the directory, then we shouldn't try to add the certificate because the certificate was already added.

gradle-jdks-setup/src/main/resources/gradle-jdks-setup.sh
```diff
@@ -58,7 +58,6 @@ do
     esac
 done
 
-APP_BASE_NAME=${0##*/}
```
crogoz, line 61: unused

gradle-jdks-test-common/src/main/groovy/com/palantir/gradle/jdks/GradleJdkTestUtils.groovy
```diff
@@ -17,12 +17,11 @@
 package com.palantir.gradle.jdks
 
 import com.palantir.gradle.jdks.setup.common.CommandRunner
-import nebula.test.IntegrationSpec
 import org.apache.commons.lang3.tuple.Pair
 
 import java.nio.file.Path
 
-abstract class GradleJdkIntegrationTest extends IntegrationSpec {
+class GradleJdkTestUtils {
```
crogoz, line 24: refactored this such that it can be used by the tests in both gradle-jdks & gradle-jdks-settings.

gradle-jdks/build.gradle
```diff
@@ -75,10 +76,10 @@ tasks.withType(JavaCompile).configureEach {
     it.options.errorprone.disable 'StrictUnusedVariable'
 }
 
-tasks.test {
-    environment("PROJECT_VERSION", project.version)
+tasks.withType(Test) {
+    environment(Map.of("PROJECT_VERSION", project.version))
     systemProperty 'ignoreDeprecations', 'true'
-    dependsOn project(':gradle-jdks-settings').tasks.named('build')
+    dependsOn project(':gradle-jdks-settings').tasks.withType(PluginUnderTestMetadata.class)
```
crogoz, line 82: before this change whenever we would run the tests in `gradle-jdks`, we would run the `gradle-jdks-settings` tests first

idea-plugin/src/main/java/com/palantir/gradle/jdks/GradleJdksProjectService.java
```diff
@@ -106,7 +107,7 @@ public void processTerminated(@NotNull ProcessEvent _event) {
                 }
             });
             consoleView.get().attachToProcess(handler);
-            ProcessTerminatedListener.attach(handler, project);
+            ProcessTerminatedListener.attach(handler, project, "Gradle JDK setup finished with exit code $EXIT_CODE$");
```
crogoz, line 110: Instead of the old "Process finished with exit code 0" log line in Intellij.

idea-plugin/src/main/java/com/palantir/gradle/jdks/GradleJdksProjectService.java
```diff
@@ -95,6 +95,7 @@ public void maybeSetupGradleJdks() {
 
     private void setupGradleJdks() {
         try {
+            consoleView.get().clear();
```
crogoz, line 98: making sure we only have the logs for the current setupGradleJdks run.

gradle-jdks-settings/src/test/groovy/com/palantir/gradle/jdks/settings/ToolchainJdksSettingsPluginTest.groovy
```diff
@@ -0,0 +1,102 @@
...
+        Files.walk(expectedJdkPath)
+                .sorted(Comparator.reverseOrder())
+                .forEach(Files::delete)
+
+        Files.walk(newInstalledJdkPath)
+                .sorted(Comparator.reverseOrder())
+                .forEach(Files::delete)
+
+        where:
+        gradleVersionNumber << [GradleJdkTestUtils.GRADLE_7_6_VERSION, GradleJdkTestUtils.GRADLE_8_5_VERSION, GradleJdkTestUtils.GRADLE_8_8_VERSION]
```
esword, line 100: why is this different than the GRADLE_TEST_VERSIONS you put in other tests?

gradle-jdks-settings/src/main/java/com/palantir/gradle/jdks/settings/ToolchainJdksSettingsPlugin.java
```diff
@@ -189,4 +201,39 @@ private static boolean isGradleVersionSupported() {
...
+    private static void runGradleJdkSetup(Path rootProjectDir) {
+        Path buildDirectory = rootProjectDir.resolve("build");
+        createDirectories(buildDirectory);
+        CommandRunner.runWithLogger(
+                new ProcessBuilder().command("./gradle/gradle-jdks-setup.sh").directory(rootProjectDir.toFile()),
+                ToolchainJdksSettingsPlugin::writeStdOutput,
+                ToolchainJdksSettingsPlugin::writeStdErr);
+    }
+
+    private static Void writeStdOutput(InputStream inputStream) {
```
esword: looks like this was copied from CaResource.  All the methods seem the same except what they do with logger.  Would it make sense to make CommandRunner non-static and put the methods in there, then pass it a logger when created?

gradle-jdks-settings/src/main/java/com/palantir/gradle/jdks/settings/ToolchainJdksSettingsPlugin.java
```diff
@@ -144,37 +149,44 @@ public Object invoke(Object _proxy, Method method, Object[] args) throws Throwab
...
-            }
+    private static List<Path> getOrInstallJdkPaths(Path rootProjectDir, Path gradleJdksLocalDirectory) {
+        List<Path> jdkPaths = getConfiguredJdkPaths(gradleJdksLocalDirectory);
+        List<Path> missingJdkPaths = getMissingPaths(jdkPaths);
+        if (!missingJdkPaths.isEmpty()) {
+            logger.error(
+                    "Gradle JDK setup is enabled (palantir.jdk.setup.enabled is true) but some jdks were not"
+                            + " installed: {}. If running from Intellij, please make sure the"
+                            + " `palantir-gradle-jdks` Intellij plugin is installed"
+                            + " https://plugins.jetbrains.com/plugin/24776-palantir-gradle-jdks/versions.",
```
CRogers: I would probably add that we're installing the JDKs manually this time (so it's obvious this is not a blocker)

gradle-jdks-setup-common/src/main/java/com/palantir/gradle/jdks/setup/common/CommandRunner.java
```diff
@@ -56,37 +46,44 @@ public static String run(List<String> commandArguments, Optional<File> directory
...
         } finally {
             executorService.shutdown();
         }
     }
 
-    public static void runWithInheritIO(List<String> commandArguments, File directory) {
-        ExecutorService executorService = Executors.newFixedThreadPool(2);
+    public static void runWithLogger(
+            ProcessBuilder processBuilder,
+            Function<InputStream, Void> stdOutputWriter,
```
CRogers: Could we use `Consumer<InputStream>` instead of `Function<InputStream, Void>`?

gradle-jdks-setup-common/src/main/java/com/palantir/gradle/jdks/setup/common/CommandRunner.java
```diff
@@ -56,37 +46,44 @@ public static String run(List<String> commandArguments, Optional<File> directory
...
+            Function<InputStream, Void> stdErrWriter) {
+        ExecutorService executorService = Executors.newFixedThreadPool(3);
```
CRogers: Do we need 3 threads here? Looks like we make only two futures?

gradle-jdks-settings/src/test/groovy/com/palantir/gradle/jdks/settings/ToolchainJdksSettingsPluginTest.groovy
crogoz: I wanted to explicitly test other gradle versions because we are using [reflection in the settings plugin](https://github.com/palantir/gradle-jdks/blob/develop/gradle-jdks-settings/src/main/java/com/palantir/gradle/jdks/settings/ToolchainJdksSettingsPlugin.java#L86-L104), so I want to make sure nothing gets broken in between gradle versions. 

pushed a [commit](https://github.com/palantir/gradle-jdks/pull/382/commits/142fb57dfe05f52c42c4105b8d4a17d446df64a4#diff-47297190c845738f2860d17927cc1cf986df22933651ea980e7345fdc840f6ecR99) that adds a comment as to why these are the gradle versions I am testing with

gradle-jdks-setup-common/src/main/java/com/palantir/gradle/jdks/setup/common/CommandRunner.java
```diff
@@ -99,5 +96,18 @@ public static String readAllInput(InputStream inputStream) {
         }
     }
 
+    public static Void processStream(InputStream inputStream, Function<String, Void> logFunction) {
```
CRogers: I think this could also be a `Consumer`?

gradle-jdks-setup-common/src/main/java/com/palantir/gradle/jdks/setup/common/CommandRunner.java
CRogers: Would avoid the `return null`s

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/GradleJdkInstallationSetup.java
CRogers: I would add a comment about this concurrency stuff.

gradle-jdks-settings/src/main/java/com/palantir/gradle/jdks/settings/ToolchainJdksSettingsPlugin.java
crogoz: I am not sure that I can pass the logger/re-use the logic.

- the Logger is different in CaResources (it is an [interface](https://github.com/palantir/gradle-jdks/blob/develop/gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/CaResources.java#L50) that can be either a `stdLogger` or a `gradle logger`). 
- The logger used here is a `gradle logger` but when I log something here (the output would be shown in Intellij), I want to always log with `.lifecycle` to make sure the logs always show up, whereas in `CaResources` the default will be `info log level`

gradle-jdks-setup-common/src/main/java/com/palantir/gradle/jdks/setup/common/CommandRunner.java
crogoz: that is perfect, changed it [here](https://github.com/palantir/gradle-jdks/pull/382/commits/5ff102c5133aaf7d5c0ba96712eca0c9e75b77ec)

CRogers:
:+1:

CRogers:
:+1:

PR #386, opened 2024-08-02
Add `/usr/java/$java_version` symlink to `install-jdks.sh` excavator script

crogoz:
## Before this PR

If autoDetection is enabled, Gradle is only searching for java executables in the directories: https://github.com/gradle/gradle/blob/80477c69ff426af47b53cda7f3167ad188935727/platforms/jvm/jvm-services/src/main/java/org/gradle/jvm/toolchain/internal/LinuxInstallationSupplier.java#L36.

## After this PR

Adding a symlink for the installed jdks in the `/usr/java` directory.

==COMMIT_MSG==
Add `/usr/java/$java_version` symlink to `install-jdks.sh` excavator script
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-386.v2.yml
+++ b/changelog/@unreleased/pr-386.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: Add `/usr/java/$java_version` symlink to `install-jdks.sh` excavator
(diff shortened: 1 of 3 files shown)


CRogers:
:+1:

PR #396, opened 2024-08-23
Write the minimum version for the `palantir-gradle-jdks` intellij plugin

crogoz:
## After this PR
The version of the `gradle-jdks` plugin will enforce the minimum version of the `palantir-gradle-jdks` Intellij plugin. This makes sure that breaking changes to the `gradle-jdks` logic will work correctly in Intellij (provided that the Intellij plugin is backwards compatible). 

This makes it possible in the future  to remove the need to update the `.gradle/config.properties` file in the `gradle-jdks` plugin (that is used by the Intellij plugin atm to set the `gradleJvm` tp `GRADLE_LOCAL_USER_HOME`) and instead have the Intellij plugin configure the gradleJvm directly to an SDK. 

==COMMIT_MSG==
Write the minimum version for the `palantir-gradle-jdks` intellij plugin
==COMMIT_MSG==

FLUPS:
- I think this logic might make sense to be moved to a common library that other idea plugins can use as well (e.g. palantir-java-format or withcraft-logging)
- Add logic to the intellij plugin to loudly fail if the current plugin version < min-version

### DIFF
--- a/gradle-jdks-groovy/src/main/groovy/com/palantir/gradle/jdks/ConfigureJdksIdeaPluginXml.groovy
+++ b/gradle-jdks-groovy/src/main/groovy/com/palantir/gradle/jdks/ConfigureJdksIdeaPluginXml.groovy
@@ -16,21 +16,57 @@
 
 package com.palantir.gradle.jdks
 
+import groovy.xml.XmlNodePrinter
+import groovy.xml.XmlParser
+import org.gradle.internal.impldep.com.google.common.collect.ImmutableMap
+import org.xml.sax.SAXException
+
+import javax.xml.parsers.ParserConfigurationException
+import java.nio.charset.StandardCharsets
+import java.nio.file.Files
+
 class ConfigureJdksIdeaPluginXml {
 
-    static void configureExternalDependencies(Node rootNode) {
-        def externalDependencies = matchOrCreateChild(rootNode, 'component', [name: 'ExternalDependencies'])
-        matchOrCreateChild(externalDependencies, 'plugin', [id: 'palantir-gradle-jdks'])
-    }
+    static void updateIdeaXmlFile(File configurationFile, String minVersion, boolean createIfAbsent) {
+        Node rootNode;
+        if (configurationFile.isFile()) {
+            try {
+                rootNode = new XmlParser().parse(configurationFile);
+            } catch (IOException | SAXException | ParserConfigurationException e) {
+                throw new RuntimeException("Couldn't parse existing configuration file: " + configurationFile, e);
+            }
+        } else {
+            if (!createIfAbsent) {
+                return;
+            }
+            rootNode = new Node(null, "project", ImmutableMap.of("version", "4"));
+        }
 
-    private static Node matchOrCreateChild(Node base, String name, Map attributes = [:], Map defaults = [:]) {
-        matchChild(base, name, attributes).orElseGet {
-            base.appendNode(name, attributes + defaults)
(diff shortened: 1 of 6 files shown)


gradle-jdks-groovy/src/main/groovy/com/palantir/gradle/jdks/ConfigureJdksIdeaPluginXml.groovy
```diff
@@ -16,21 +16,57 @@
...
 
-    private static Optional<Node> matchChild(Node base, String name, Map attributes = [:]) {
-        def child = base[name].find { it.attributes().entrySet().containsAll(attributes.entrySet()) }
-        return Optional.ofNullable(child)
+    static void configureExternalDependencies(Node rootNode, String minVersion) {
+        def externalDependencies = matchOrCreateChild(rootNode, 'component', [name: 'ExternalDependencies'])
+        matchOrCreateChild(externalDependencies, 'plugin', [ 'id': 'palantir-gradle-jdks' ], ['min-version' : minVersion])
+    }
+
+    private static Node matchOrCreateChild(Node base, String name, Map keyAttributes = [:], Map otherAttributes = [:]) {
```
crogoz, line 62: different from the rest of the intellij externalDependencies.xml updater logic ([witchcraft-logging logic](https://github.com/palantir/witchcraft-java-logging/blob/9369f35698cdf0e9e13b4671dfe68ab57b8846f6/gradle-witchcraft-logging/src/main/groovy/com/palantir/witchcraft/java/logging/gradle/idea/IdeaPluginRecommenderPlugin.groovy#L49) and [palantir-java-format logic](https://github.com/palantir/palantir-java-format/blob/6809942df899297f319bfd2d5085d2a9d31dda8c/gradle-palantir-java-format/src/main/groovy/com/palantir/javaformat/gradle/ConfigureJavaFormatterXml.groovy#L34)). The other plugins would add a new entry for the new min-version, instead of updating it.

CRogers:
I would add a bit more context in the PR Description if nothing else to help future version of ourselves  - why exactly does this make it easier to change the intellij plugin later?

gradle-jdks-groovy/src/main/groovy/com/palantir/gradle/jdks/ConfigureJdksIdeaPluginXml.groovy
```diff
@@ -16,21 +16,57 @@
...
+
+        try (BufferedWriter writer = Files.newBufferedWriter(configurationFile.toPath(), StandardCharsets.UTF_8);
+             PrintWriter printWriter = new PrintWriter(writer)) {
+            XmlNodePrinter nodePrinter = new XmlNodePrinter(printWriter);
+            nodePrinter.setPreserveWhitespace(true);
+            nodePrinter.print(rootNode);
+        } catch (IOException e) {
+            throw new RuntimeException("Failed to write back to configuration file: " + configurationFile, e);
         }
     }
```
CRogers, line 55: This is all moved basically as in from the other file right?

CRogers (approved):

gradle-jdks-groovy/src/main/groovy/com/palantir/gradle/jdks/ConfigureJdksIdeaPluginXml.groovy
crogoz: yes `updateIdeaXmlFile` is moved with no changes. 

`configureExternalDependencies ` & `matchOrCreateChild ` are both changed

PR #403, opened 2024-09-04
[Gradle JDK Automanagement] Adding all systemCerts by default in the installed JDKs

crogoz:
## Before this PR
The JDK installation directory contained in the name the hash of the certificates that were required (from the `certs/` directory). This is not really accurate. 

We could have a project that would require the Palantir cert (having the `certs/Palantir.serial-number` file) but if the user (eg. external users, opensource CircleCi) that runs the project doesn't have the Palantir certificate in the systemCertificate trustore, then we wouldn't import the certificate(https://github.com/palantir/gradle-jdks/pull/403/files#diff-d2a5b341108734345bf72bfb7dcb34cdc12976f8ffcf2b99df905702ce8faa12L174) and the hash would not be accurate. That means that the `checkGradleJdkConfigs` task would fail here: https://github.com/palantir/gradle-jdks/blob/develop/gradle-jdks/src/main/java/com/palantir/gradle/jdks/CheckGradleJdksConfigsTask.java#L51C9-L51C26 because the hash would be different for the external user/opensource Circle.

## After this PR
Note: the changes only affect the Gradle JDK Automanagement workflow.
- all systemCerts will be added by default to the installed JDKs
- the installation directory won't contain the hash anymore
- we won't generate the `certs/` directory anymore

==COMMIT_MSG==
[Gradle JDK Automanagement] Adding all systemCerts by default in the installed JDKs
==COMMIT_MSG==

Possible FLUPS: if the system certificates are renewed, we would need to re-install them in the JDKs. Given that we are constantly bumping the jdks versions (using the `jdks.latest` plugin), we won't do the check if a JDK needs to update the certificates to be up-to-date with the system certs.


gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/CaResources.java
```diff
@@ -53,113 +62,73 @@ public CaResources(ILogger logger) {
...
-                    "-keystore",
-                    jdkInstallationDirectory.resolve("lib/security/cacerts").toString());
-            CommandRunner.runWithOutputCollection(new ProcessBuilder().command(checkCertificateCommand));
-            return true;
-        } catch (Exception e) {
-            if (e.getMessage().contains(String.format("Alias <%s> does not exist", alias))) {
-                return false;
+            LdapName ldapName = new LdapName(distinguishedName);
+            for (Rdn rdn : ldapName.getRdns()) {
+                if ("CN".equalsIgnoreCase(rdn.getType())) {
```
crogoz, line 168: I couldn't find a better way to create an alias. 
The aliases would look like: `gradlejdks_palantir3rd-generationrootca` if the CN is found, otherwise something like: `GradleJdks_OU=GoDaddyClass2CertificationAuthority,O=TheGoDaddyGroup\,Inc.,C=US`

gradle-jdks-excavator-configurations/src/main/java/com/palantir/gradle/jdks/GradleJdksConfigurator.java
```diff
@@ -68,28 +61,18 @@ private static void writeJdkInstallationConfiguration(
...
         Path jdksDir = targetDir.resolve("jdks");
         Path jdkOsArchDir = jdksDir.resolve(javaVersion).resolve(os.uiName()).resolve(arch.uiName());
         try {
             Files.createDirectories(jdkOsArchDir);
             GradleJdksConfigsUtils.writeConfigurationFile(
-                    jdkOsArchDir.resolve("download-url"), resolveDownloadUrl(baseUrl, jdkSpec));
+                    jdkOsArchDir.resolve("download-url"), resolveDownloadUrl(baseUrl, jdkDistributionName, jdkRelease));
             GradleJdksConfigsUtils.writeConfigurationFile(
-                    jdkOsArchDir.resolve("local-path"), resolveLocalPath(jdkSpec));
+                    jdkOsArchDir.resolve("local-path"), resolveLocalPath(jdkDistributionName, jdkRelease));
```
CRogers, line 75: So we are keeping the local path? I think this may make sense tbh.

crogoz: Yes, I also think this might be better. It would have ended up a little bit too complicated to extract the local-path in bash given that the downloadUrls might have had multiple dots also in the name of the file and the name of the extensions (tar.bz2 etc. )

gradle-jdks/src/main/java/com/palantir/gradle/jdks/GenerateGradleJdksConfigsTask.java
```diff
@@ -37,6 +39,20 @@ protected final Directory gradleDirectory() {
...
+    protected final void maybePrepareForAction(List<Path> targetPaths) {
+        for (Path targetPath : targetPaths) {
+            if (!Files.exists(targetPath)) {
+                continue;
+            }
+            if (Files.isDirectory(targetPath)) {
+                FileUtils.deleteDirectory(targetPath);
+            } else {
+                FileUtils.deleteFile(targetPath);
+            }
```
CRogers: Maybe extract this into `FileUtils`?

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/CaResources.java
```diff
@@ -53,113 +62,73 @@ public CaResources(ILogger logger) {
...
+    }
+
+    private Optional<KeyStore> loadKeystore(char[] password, Path location) {
+        try (InputStream keystoreStream = new BufferedInputStream(Files.newInputStream(location))) {
+            KeyStore keystore = KeyStore.getInstance("JKS");
+            keystore.load(keystoreStream, password);
+            return Optional.of(keystore);
+        } catch (KeyStoreException | CertificateException | NoSuchAlgorithmException | IOException e) {
+            logger.log(String.format("Couldn't load jks, an exception occurred %s", e));
+            return Optional.empty();
```
CRogers: I'd just have this throw here rather than returning an optional, losing information, then rethrowing via `orElseThrow`. This way we can keep whole stacktrace.

CRogers: I'd also add the location to the error message.

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/CaResources.java
```diff
@@ -53,113 +62,73 @@ public CaResources(ILogger logger) {
...
-    }
+            jks.store(new BufferedOutputStream(new FileOutputStream(jksPath.toFile())), passwd);
 
-    private void writeStdError(InputStream inputStream) {
-        CommandRunner.processStream(inputStream, logger::logError);
+            logger.log("Certificates imported successfully into the KeyStore.");
+        } catch (KeyStoreException | FileNotFoundException e) {
+            throw new RuntimeException("Failed to import certificates", e);
+        } catch (CertificateException | IOException | NoSuchAlgorithmException e) {
+            throw new RuntimeException(e);
```
CRogers: I'd just unify all these catches into one with the "Failed to import certificates" message

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/CaResources.java
```diff
@@ -53,113 +62,73 @@ public CaResources(ILogger logger) {
...
-            if (isCertificateInTruststore(jdkInstallationDirectory, aliasContentCert.getAlias())) {
-                logger.log("Certificate already exists in the truststore, skipping...");
-                return;
+            char[] passwd = "changeit".toCharArray();
+            Path jksPath = jdkInstallationDirectory.resolve("lib/security/cacerts");
+            KeyStore jks =
+                    loadKeystore(passwd, jksPath).orElseThrow(() -> new RuntimeException("Could not load keystore"));
+
+            for (Certificate cert : certificates) {
+                String alias = getAlias((X509Certificate) cert);
```
CRogers: Is it worth at least trying to not insert the system cert if it already exists in the truststore? In the interest of making as few changes a possible at least?

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/CaResources.java
CRogers: I wonder if there are any characters that are not allowed in the alias?

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/CaResources.java
crogoz: that is if the truststore has duplicated certificates ?

crogoz: the system certificates will only be added if that jdk was just installed: https://github.com/palantir/gradle-jdks/pull/403/files#diff-d2a5b341108734345bf72bfb7dcb34cdc12976f8ffcf2b99df905702ce8faa12R110

CRogers: I expect a lot of the certificates in system truststore are already in the jdk truststore.

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/CaResources.java
```diff
@@ -74,20 +75,44 @@ private void importCertificates(Path jdkInstallationDirectory, List<Certificate>
...
+    private static Set<BigInteger> getExistingCertificates(KeyStore keyStore) {
+        try {
+            return Collections.list(keyStore.aliases()).stream()
+                    .map(alias -> {
+                        try {
+                            return Optional.ofNullable(keyStore.getCertificate(alias))
+                                    .map(X509Certificate.class::cast);
+                        } catch (KeyStoreException e) {
+                            throw new RuntimeException(e);
+                        } catch (ClassCastException e) {
```
CRogers: Rather than catching the exception, can we do `.filter(X509Certificate.class::isAssignableFrom)` before the `map(cast)`?

gradle-jdks/src/main/java/com/palantir/gradle/jdks/GradleJdksConfigs.java
```diff
@@ -63,17 +63,26 @@ protected abstract void applyGradleJdkFileAction(
...
     public final void action() {
+        Path gradleJdksDir = gradleDirectory().dir("jdks").getAsFile().toPath();
+        Path gradleJdksSetupJar =
+                gradleDirectory().file(GRADLE_JDKS_SETUP_JAR).getAsFile().toPath();
+        Path gradleJdksFunctionsScript =
+                gradleDirectory().file(GRADLE_JDKS_FUNCTIONS_SCRIPT).getAsFile().toPath();
+        Path gradleJdksSetupScript =
+                gradleDirectory().file(GRADLE_JDKS_SETUP_SCRIPT).getAsFile().toPath();
+        // Older versions of `gradle.jdks` would also populate the `certs/` directory
+        // TODO(crogoz): Remove this once everyone is on the newer `com.palantir.jdks`
```
CRogers, line 78: Good thinking!

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/CaResources.java
```diff
@@ -53,113 +62,93 @@ public CaResources(ILogger logger) {
...
+            KeyStore jks = loadKeystore(passwd, jksPath);
+            Set<BigInteger> existingSerialNumbers = getExistingCertificates(jks);
+            logger.log(String.format("Certificate %s imported successfully into the KeyStore.", existingSerialNumbers));
+            List<X509Certificate> newCertificates = certificates.stream()
+                    .map(X509Certificate.class::cast)
+                    .filter(certificate -> !existingSerialNumbers.contains(certificate.getSerialNumber()))
+                    .collect(Collectors.toList());
+            for (X509Certificate certificate : newCertificates) {
+                String alias = getAlias(certificate);
+                logger.log(String.format("Certificate %s imported successfully into the KeyStore.", alias));
```
CRogers: I just tried this branch on my laptop, and I got some interesting logging. First, I got a big long line of numbers, which I don't know what it is:

Screenshot 2024-09-06 at 13 31 35

I also ended up with multiple certs being imported with the same alias:

```
Certificate GradleJdks_computerName imported successfully into the KeyStore.
Certificate GradleJdks_computerName imported successfully into the KeyStore.
```

I think the first of these are being overwritten by the second (I assume they are actually different serial numbers - haven't checked). We can probably fix this by putting the serial number into the alias too?

gradle-jdks-excavator-configurations/src/main/java/com/palantir/gradle/jdks/GradleJdksConfigurator.java
```diff
@@ -110,13 +93,13 @@ private static void writeInstallationScripts(Path targetDir) {
         GradleJdksConfigsUtils.copyResourceToPath(scriptsDir, "gradle-jdks-setup.jar");
     }
 
-    private static String resolveLocalPath(JdkSpec jdkSpec) {
-        return String.format(
-                "%s-%s-%s", jdkSpec.distributionName(), jdkSpec.release().version(), jdkSpec.consistentShortHash());
+    private static String resolveLocalPath(JdkDistributionName jdkDistributionName, JdkRelease jdkRelease) {
+        return String.format("%s-%s", jdkDistributionName, jdkRelease.version());
```
CRogers, line 97: I wonder, if for absolute safety, we should include the os name and arch in the local path. These were being included in the hash before (now gone).

It might seem like this is unnecessary as the OS ~~or arch~~ does not change; but that's not necessarily the case. 

1. ~~You can run programs as if they were x86-64 in macos using the `arch` command~~ This doesn't do what I thought it did
2. On windows (if we ever support it), if you run normally once it will install windows jdks here, then if you load the project with WSL it will still use the windows versions even though you are "running linux". I don't actually know how bad this is, it might actually still just work.

CRogers: Eh, on seconds thoughts this is probably fine. If it ever does become an issue, we can just add the os and/or arch and it'll work out.

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/CaResources.java
felixdesouza: If you use streamex, it becomes even nicer:

```java
try {
  StreamEx.of(keyStore.aliases())
    // might need to extract to a method or inline inside, but it's minimal
    .map(keystore::getCertificate) 
    .nonNull()
    .select(X509Certificate.class)
    .map(X509Certificate::getSerialNumber)
    .toSet();
} catch (KeyStoreException e) {
  throw new RuntimeException(e);
}
```

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/CaResources.java
```diff
@@ -53,113 +62,93 @@ public CaResources(ILogger logger) {
...
-    private void unixImportCertInJdk(AliasContentCert aliasContentCert, Path jdkInstallationDirectory) {
+    private void importCertificates(Path jdkInstallationDirectory, List<Certificate> certificates) {
         try {
-            if (isCertificateInTruststore(jdkInstallationDirectory, aliasContentCert.getAlias())) {
-                logger.log("Certificate already exists in the truststore, skipping...");
-                return;
+            char[] passwd = "changeit".toCharArray();
+            Path jksPath = jdkInstallationDirectory.resolve("lib/security/cacerts");
+            KeyStore jks = loadKeystore(passwd, jksPath);
+            Set<BigInteger> existingSerialNumbers = getExistingCertificates(jks);
```
CRogers: Serial number is actually bad here. I had a look in my system truststore, and there are 8 certs with serial number `0`, 6 with serial number `1` etc:

Screenshot 2024-09-06 at 14 10 12

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/CaResources.java
CRogers: I think the `[0, 1, 2` is a list of serial numbers somehow.

crogoz: yes, I think I added the wrong log line here: https://github.com/palantir/gradle-jdks/pull/403/files/4b6a87a4346fae20fb06dbbd67bc740850013f74#diff-f0333751b6ca40fe3d118c9f2fce18e089d40d62e5bb0649a7a0635b82e5df56R79. Deleting it

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/CaResources.java
CRogers: We probably want to use some hash of the certificate text somehow... there's `X509Certificate#getSignature` but it returns a byte[] - could probably turn that into a hex string?

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/CaResources.java
```diff
@@ -53,113 +62,93 @@ public CaResources(ILogger logger) {
...
+            logger.log(String.format("Certificate %s imported successfully into the KeyStore.", existingSerialNumbers));
+            List<X509Certificate> newCertificates = certificates.stream()
+                    .map(X509Certificate.class::cast)
```
felixdesouza: is this a safe cast?

Probably can do the same thing with the `StreamEx#select` from `getExistingCertificates`

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/CaResources.java
```diff
@@ -53,113 +62,93 @@ public CaResources(ILogger logger) {
...
+                    .filter(certificate -> !existingSerialNumbers.contains(certificate.getSerialNumber()))
```
felixdesouza: up to you, but you can do 

    StreamEx#remove(certificate -> existingSerialNumbers.contains(certificate.getSerialNumber()))

Difference being the intent is a tiny bit clearer, but it's mostly syntactic so not blocking

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/CaResources.java
```diff
@@ -53,113 +62,93 @@ public CaResources(ILogger logger) {
...
+        return selectPalantirCertificate(systemCertificates());
     }
 
-    public Optional<AliasContentCert> maybeGetCertificateFromSerialNumber(String serialNumber, String alias) {
-        return systemCertificates()
-                .map(certs -> selectCertificates(certs, Map.of(serialNumber, alias)))
-                .orElseGet(Stream::empty)
-                .findFirst();
+    public void importAllSystemCerts(Path jdkInstallationDirectory) {
+        importCertificates(jdkInstallationDirectory, parseCerts(systemCertificates()));
```
felixdesouza, line 73: why not have `parseCerts` return you a list of `X509Certificate` ?

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/CaResources.java
crogoz: I think the `<CA, serialNumber>` pairs should be unique. I'll try to change the alias to account for this. 
I think the 0 serial number might be indeed invalid. Maybe a revoked certificate ? I will try to see if I can check the validity of the serialNumber somehow

crogoz: Screenshot 2024-09-06 at 3 17 28 PM

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/CaResources.java
felixdesouza: if you have the `certificates` argument be `Set<X509Certificate>`, you can just make this be a `Sets.difference` call instead?

Then `getExistingCertificates` returns `Set<X509Certificate` as well

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/CaResources.java
felixdesouza: see below, but can you just use the object equality for checking existence? then for logging, just use the serial number or the fingerprint

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/CaResources.java
```diff
@@ -53,113 +62,93 @@ public CaResources(ILogger logger) {
...
         try {
-            if (isCertificateInTruststore(jdkInstallationDirectory, aliasContentCert.getAlias())) {
-                logger.log("Certificate already exists in the truststore, skipping...");
-                return;
+            char[] passwd = "changeit".toCharArray();
+            Path jksPath = jdkInstallationDirectory.resolve("lib/security/cacerts");
+            KeyStore jks = loadKeystore(passwd, jksPath);
+            Set<BigInteger> existingSerialNumbers = getExistingCertificates(jks);
+            logger.log(String.format("Certificate %s imported successfully into the KeyStore.", existingSerialNumbers));
+            List<X509Certificate> newCertificates = certificates.stream()
```
CRogers, line 82: There are multiple types of `X509Certificate`s. From the output it looks like we're including code signing certs, certs used for kerberos, wifi EAP etc. We only really need the ones used for TLS, and probably only the ones that are CAs at that.

This is kinda tricky though. I think we can check this by looking at `getKeyUsage()` which I think should have both `Digital Signature` and `Key Encipherment`, with extended usage of at least `TLS Web Server Authentication`. You can then see if `CA` is true in the basic constraints. The API for inspecting these are all _terrible_ though. Also, some of these very old CA certs (the Starfield one for example) do not have key usage; I think for these you could just import them.

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/CaResources.java
CRogers: The Starfield certs have 0 serial number and is valid until the 2030s. One at least is also a root cert so has no CA. I think serial number is basically garbage here - we need to use the signature/fingerprint.

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/CaResources.java
CRogers: It looks like `Certificate` implements hashCode and equals using the encoded form of the certificates, so it should work and means we don't need to worry about signature/fingerprints?

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/CaResources.java
CRogers: I'd say "...imported successfully into the JDK truststore from the system truststore" explicitly so people know where it came from.

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/CaResources.java
crogoz: I haven't used StreamEx because CaResources will be included in the `gradle-jdks-setup` fat jar that will be shipped to all repos, so I am trying to keep the jar dependencies as small as possible: https://github.com/palantir/gradle-jdks/blob/develop/gradle-jdks-setup/build.gradle#L11

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/CaResources.java
crogoz: > is this a safe cast?

good point, I changed in: https://github.com/palantir/gradle-jdks/pull/403/commits/902534cb58e187c454a8a386330fc0c8850922b2

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/CaResources.java
```diff
@@ -53,121 +65,144 @@ public CaResources(ILogger logger) {
...
+    private static boolean hasCaCertUsage(X509Certificate certificate) {
+        boolean[] keyUsage = certificate.getKeyUsage();
+        if (keyUsage == null) {
             return true;
-        } catch (Exception e) {
-            if (e.getMessage().contains(String.format("Alias <%s> does not exist", alias))) {
-                return false;
+        }
+        // digitalSignature and keyEncipherment are enabled
+        if (keyUsage[0] && keyUsage[2]) {
```
CRogers, line 138: I'd make the 0 and 2 (and 5) here constants so we can give them a name like `DIGITAL_SIGNATURE` so it looks like `keyUsage[DIGITAL_SIGNATURE]` and is a bit more self documenting.

CRogers (approved):

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/CaResources.java
CRogers: oops, forgot I had suggested this. No worries.

Issue #394, opened 2024-08-21
IncorrectOperationException thrown by intellij Plugin


fawind:
## What happened?

Intellij reports the following "IDE Internal Error" that's caused by the gradle-jdks intellij plugin.

I believe this happens when the Intellij embedded terminal window is in focus while the below code runs.

<details>
<summary>Stacktrace:</summary>

```
com.intellij.util.IncorrectOperationException: Sorry but parent: com.intellij.execution.impl.ConsoleViewImpl[,0,0,2560x292,invalid,layout=java.awt.BorderLayout,alignmentX=0.0,alignmentY=0.0,border=,flags=9,maximumSize=,minimumSize=,preferredSize=] (class com.intellij.execution.impl.ConsoleViewImpl) has already been disposed (see the cause for stacktrace) so the child: com.intellij.util.Alarm@6683437c (class com.intellij.util.Alarm) will never be disposed
	at com.intellij.openapi.util.ObjectTree.register(ObjectTree.java:43)
	at com.intellij.openapi.util.Disposer.register(Disposer.java:161)
	at com.intellij.util.Alarm.<init>(Alarm.java:145)
	at com.intellij.execution.impl.ProcessStreamsSynchronizer.<init>(ProcessStreamsSynchronizer.java:52)
	at com.intellij.execution.impl.ConsoleViewRunningState.<init>(ConsoleViewRunningState.java:49)
	at com.intellij.execution.impl.ConsoleViewImpl$1.attachTo(ConsoleViewImpl.java:167)
	at com.intellij.execution.impl.ConsoleViewImpl.attachToProcess(ConsoleViewImpl.java:313)
	at com.palantir.gradle.jdks.GradleJdksProjectService.setupGradleJdks(GradleJdksProjectService.java:108)
	at com.palantir.gradle.jdks.GradleJdksProjectService.maybeSetupGradleJdks(GradleJdksProjectService.java:93)
	at com.palantir.gradle.jdks.GradleJdksExternalSystemTaskNotificationListener.onStart(GradleJdksExternalSystemTaskNotificationListener.java:33)
	at com.intellij.openapi.externalSystem.service.remote.ExternalSystemProgressNotificationManagerImpl$onStart$1.invoke(ExternalSystemProgressNotificationManagerImpl.kt:39)
[...]

PR #405, opened 2024-09-05
[Gradle JDK Automanagement][Intellij plugin] Background tool window

### LINKED ISSUE #395: Intellij plugin: Can the popup not steal focus
## What happened?
When it runs, the window steals focus which can be quite jarring + the window opened is quite large.

## What did you want to happen?
Run in the background like the normal gradle import action and only steal focus when it fails.

See internal reproducer gif

crogoz:
## After this PR
Fixes: https://github.com/palantir/gradle-jdks/issues/395
- `Gradle JDK setup` tool window will be minimized by default
- The `Gradle JDK setup` would look as if it were a background process (similar to the `Gradle Sync` workflow)
- If the `Gradle JDK setup` fails, then we will show a pop-up Notification with a link to the logs. 

==COMMIT_MSG==
[Gradle JDK Automanagement][Intellij plugin] Background tool window
==COMMIT_MSG==

Note: I've tried using the declarative setup for the `Gradle JDK setup` toolWindow (https://plugins.jetbrains.com/docs/intellij/tool-windows.html#declarative-setup) but the workflow had 2 issues: 
-  some initial logs might be missing from the consoleView, I suspect this is because the [consoleView creation](https://github.com/palantir/gradle-jdks/blob/f529abb2eaa2e5ae1d477906036fabec49d51260/idea-plugin/src/main/java/com/palantir/gradle/jdks/GradleJdkToolWindowFactory.java#L43) would happen after we started [running the gradle jdk setup](https://github.com/palantir/gradle-jdks/blob/f529abb2eaa2e5ae1d477906036fabec49d51260/idea-plugin/src/main/java/com/palantir/gradle/jdks/GradleJdksProjectService.java#L120C13-L120C29)


idea-plugin/src/main/java/com/palantir/gradle/jdks/GradleJdksProjectService.java
```diff
@@ -90,7 +110,28 @@ public void maybeSetupGradleJdks() {
                     "Skipping setupGradleJdks because gradle JDK setup is not found %s", gradleSetupScript));
             return;
         }
-        setupGradleJdks();
+        TasksKt.withBackgroundProgress(
```
crogoz, line 111: instead of: https://github.com/JetBrains/intellij-community/blob/99964dd3c76f6cf9eb24018c002a68848e54dddb/platform/core-api/src/com/intellij/openapi/progress/Task.java#L43C8-L43C12 which has obsolescence notice

idea-plugin/src/main/java/com/palantir/gradle/jdks/GradleJdksProjectService.java
```diff
@@ -90,7 +110,28 @@ public void maybeSetupGradleJdks() {
...
+                },
+                new Continuation<>() {
+                    @Override
+                    public @NotNull CoroutineContext getContext() {
+                        return EmptyCoroutineContext.INSTANCE;
+                    }
+
+                    @Override
+                    public void resumeWith(@NotNull Object _object) {}
+                });
```
crogoz, line 132: the Gradle JDK setup process would look similar to the Gradle Sync background process. The logs can be seen in `Gradle JDK Setup` window tab. 

Screenshot 2024-09-06 at 12 21 50 PM

idea-plugin/src/main/java/com/palantir/gradle/jdks/GradleJdksProjectService.java
```diff
@@ -109,6 +151,21 @@ public void processTerminated(@NotNull ProcessEvent _event) {
             consoleView.get().attachToProcess(handler);
             ProcessTerminatedListener.attach(handler, project, "Gradle JDK setup finished with exit code $EXIT_CODE$");
             handler.waitFor();
+            if (handler.getExitCode() != 0) {
+                Notification notification = NotificationGroupManager.getInstance()
```
crogoz, line 153: if the process has failed for any reason, a pop-up will appear: 

Screenshot 2024-09-06 at 12 28 17 PM

I could also focus the `Gradle JDK setup` window tool by default. wdyt ?

idea-plugin/src/main/java/com/palantir/gradle/jdks/GradleJdksProjectService.java
```diff
@@ -72,13 +83,22 @@ private ConsoleView initConsoleView() {
             ContentFactory contentFactory = ContentFactory.getInstance();
             Content content = contentFactory.createContent(newConsoleView.getComponent(), "", false);
             toolWindow.getContentManager().addContent(content);
-            // TODO(crogoz): Focus only when error or takes a long time?
-            toolWindow.activate(null);
+            Disposer.register(project, newConsoleView);
```
crogoz, line 84: maybe this fixes: https://github.com/palantir/gradle-jdks/issues/394 I wasn't able to repro the issue

idea-plugin/src/main/java/com/palantir/gradle/jdks/GradleJdksProjectService.java
felixdesouza: I think it makes sense to focus the window on failure. The notification windows are quite quiet. I for some reason barely ever notice the "required plugins" notification lolol and I think it's quite common.

idea-plugin/src/main/java/com/palantir/gradle/jdks/GradleJdksProjectService.java
felixdesouza: 🙏

idea-plugin/src/main/java/com/palantir/gradle/jdks/GradleJdksProjectService.java
crogoz: addressed in: https://github.com/palantir/gradle-jdks/pull/405/commits/aceac47034cba90dacecf8fe3f41de79337b5331

CRogers (approved):

PR #408, opened 2024-09-11
Use com.palantir.external-publish-intellij for idea plugins

FinlayRJW:
## Before this PR
Had the same copy and pasted block of code for publishing intellij plugins which was not centrally managed

## After this PR
Migrated to the `com.palantir.external-publish-intellij` plugin to handle the publication of intellij plugins

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,7 +10,7 @@ buildscript {
         classpath 'com.palantir.gradle.jdks:gradle-jdks:0.46.0'
         classpath 'com.palantir.gradle.jdkslatest:gradle-jdks-latest:0.15.0'
         classpath 'org.unbroken-dome.gradle-plugins:gradle-testsets-plugin:4.1.0'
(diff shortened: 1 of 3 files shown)


FinlayRJW:
Need to update the version of `com.palantir.gradle.externalpublish:gradle-external-publish-plugin` before merging

build.gradle
```diff
@@ -22,7 +22,6 @@ buildscript {
 }
 
 plugins {
-    id "org.jetbrains.intellij" version "1.17.3" apply false
     id 'org.jetbrains.gradle.plugin.idea-ext' version "1.1.1"
```
FinlayRJW: Do you think this can be removed - the checks still pass without it but don't want to remove it if its needed somewhere

CRogers: I don't think we actually use this, can probably be removed.

FinlayRJW: awesome!

CRogers (approved):

PR #410, opened 2024-09-11
[Intellij plugin] Prompt user to update plugin when installed version is lower than minVersion

crogoz:
## Before this PR

Users would only get a notification in Intellij when the minVersion specified in externalDependencies.xml file is lower than the installed plugin version. 
 
Screenshot 2024-09-11 at 4 57 26 PM

## After this PR
For the `palantir-gradle-jdks` plugin only, we will check first (before running the gradle jdks setup) if the plugin meets the required minimum version. If not, then it will show a notification and it will show the Settings > Plugins window
Note: the check will run when sync'ing the gradle projects or when running gradle tasks (similar to the gradle jdk setup). 

https://github.com/user-attachments/assets/4b6b5732-fe05-4e65-9634-67bc9bda0b98

Notes: 
intellij provides a plugin checker: https://github.com/JetBrains/intellij-community/blob/master/platform/platform-impl/src/com/intellij/externalDependencies/impl/CheckRequiredPluginsActivity.java but it doesn't seem to have the option to update plugins (it just prints the error message).

==COMMIT_MSG==
[Intellij plugin] Prompt user to update plugin when installed version is lower than minVersion
==COMMIT_MSG==


idea-plugin/src/main/java/com/palantir/gradle/jdks/PluginUpdateCheckerService.java
```diff
@@ -0,0 +1,77 @@
...
+            Notification notification = NotificationGroupManager.getInstance()
+                    .getNotificationGroup("Update Palantir plugins")
+                    .createNotification(
+                            "Update palantir-gradle-jdks plugin",
+                            String.format(
+                                    "Please update the plugin in the Settings window to a version higher than '%s'",
+                                    maybeMinVersion.get()),
+                            NotificationType.ERROR);
+            notification.notify(project);
+            PluginsAdvertiser.installAndEnablePlugins(Set.of(PLUGIN_ID), notification::expire);
```
crogoz: unfortunately I couldn't test this properly when running `./gradlew runIde` since the new Intellij instance has already the latest versions of all the plugins installed. I do expect to see a similar window to (see below the "Update" button) if there is a newer version of the plugin available.
Screenshot 2024-09-12 at 1 47 52 PM

idea-plugin/src/main/resources/META-INF/plugin.xml
```diff
@@ -1,4 +1,4 @@
-<idea-plugin url="https://github.com/palantir/gradle-jdks">
+<idea-plugin url="https://github.com/palantir/gradle-jdks" require-restart="false">
```
crogoz: following: https://plugins.jetbrains.com/docs/intellij/dynamic-plugins.html

Ran: Plugin DevKit | Plugin descriptor | Plugin.xml dynamic plugin verification  from https://plugins.jetbrains.com/docs/intellij/dynamic-plugins.html#restrictions

idea-plugin/src/main/java/com/palantir/gradle/jdks/PluginUpdateCheckerService.java
```diff
@@ -0,0 +1,77 @@
...
+        if (!isPluginUpToDate) {
+            Notification notification = NotificationGroupManager.getInstance()
+                    .getNotificationGroup("Update Palantir plugins")
+                    .createNotification(
+                            "Update palantir-gradle-jdks plugin",
+                            String.format(
+                                    "Please update the plugin in the Settings window to a version higher than '%s'",
+                                    maybeMinVersion.get()),
+                            NotificationType.ERROR);
+            notification.notify(project);
```
CRogers: From the video it looks like if you click the notification it takes you to the update page. But I don't think that's super obvious it's going to do that. Can we add a link or button like the install extra plugins button?

Screenshot 2024-09-12 at 12 56 23

idea-plugin/src/main/java/com/palantir/gradle/jdks/PluginUpdateCheckerService.java
```diff
@@ -0,0 +1,77 @@
...
+                        .filter(dependencyOnPlugin ->
+                                dependencyOnPlugin.getPluginId().equals(pluginId.getIdString()))
+                        .map(DependencyOnPlugin::getMinVersion)
+                        .filter(version -> Optional.ofNullable(version).isPresent())
+                        .max(VersionComparatorUtil::compare);
+        boolean isPluginUpToDate = maybeMinVersion
+                .map(minVersion -> VersionComparatorUtil.compare(pluginDescriptor.getVersion(), minVersion) > 0)
+                .orElse(true);
+
+        if (!isPluginUpToDate) {
```
CRogers: Would probably just invert this if to return early.

idea-plugin/src/main/java/com/palantir/gradle/jdks/PluginUpdateCheckerService.java
```diff
@@ -0,0 +1,77 @@
...
+                            "Update palantir-gradle-jdks plugin",
+                            String.format(
+                                    "Please update the plugin in the Settings window to a version higher than '%s'",
+                                    maybeMinVersion.get()),
+                            NotificationType.ERROR);
+            notification.notify(project);
+            PluginsAdvertiser.installAndEnablePlugins(Set.of(PLUGIN_ID), notification::expire);
+        }
+    }
+}
```
CRogers, line 78: Do we still get the other not so useful (see below) notification? Or is it suppressed somehow.

Screenshot 2024-09-06 at 16 44 27

idea-plugin/src/main/java/com/palantir/gradle/jdks/PluginUpdateCheckerService.java
crogoz: > From the video it looks like if you click the notification it takes you to the update page. 

It opens up both the notification and the window to install the plugins. In the video I was just trying to highlight that there is also a pop-up that shows up, but I didn't click on it. 

Do you think it is better to just have the pop-up with an "Update" button and if the user clicks on it to actually open the Installation window ?

idea-plugin/src/main/java/com/palantir/gradle/jdks/PluginUpdateCheckerService.java
crogoz: Yes, we still get it, we'd get both. I can remove my notification, but I thought it might be nice to actually tell you that the you need to use the pop-up window to update the plugin.

idea-plugin/src/main/java/com/palantir/gradle/jdks/PluginUpdateCheckerService.java
CRogers: I think it's just best to put the update settings screen in people's faces - I think what you've done here is probably best, I just didn't fully understand from the video.

idea-plugin/src/main/java/com/palantir/gradle/jdks/GradleJdksExternalSystemTaskNotificationListener.java
```diff
@@ -33,6 +33,7 @@ public void onStart(@NotNull ExternalSystemTaskId id, String _workingDir) {
                         || id.getType() == ExternalSystemTaskType.EXECUTE_TASK)) {
             Project project = id.findProject();
             if (project != null) {
+                project.getService(PluginUpdateCheckerService.class).checkPluginVersion();
```
CRogers, line 36: I guess when projects are being refreshed _is_ the place to do this as that's when the `.idea/external-dependencies.xml` is being changed.

changelog/@unreleased/pr-410.v2.yml
```diff
@@ -0,0 +1,6 @@
+type: improvement
+improvement:
+  description: '[Intellij plugin] Prompt user to update plugin when installed version
```
CRogers, line 3: I guess this is fine to try out in the this project for now; maybe eventually we can either make it a library each plugin uses (might be bad if they all try to open the update window at once though?) or a separate plugin that updates all the other plugins?

CRogers:
:+1:

idea-plugin/src/main/java/com/palantir/gradle/jdks/PluginUpdateCheckerService.java
```diff
@@ -0,0 +1,77 @@
...
+        IdeaPluginDescriptor pluginDescriptor = PluginManagerCore.getPlugin(pluginId);
+        if (pluginDescriptor == null) {
+            logger.info("Plugin " + pluginId + " not found");
+            return;
+        }
+        Optional<String> maybeMinVersion =
+                ExternalDependenciesManager.getInstance(project).getDependencies(DependencyOnPlugin.class).stream()
+                        .filter(dependencyOnPlugin ->
+                                dependencyOnPlugin.getPluginId().equals(pluginId.getIdString()))
+                        .map(DependencyOnPlugin::getMinVersion)
```
CRogers, line 58: Just thought of a possible complication:

1. We release a new version of a gradle + intellij plugin, set the min-version of intellij plugin to "force" people to upgrade.
2. The Gradle plugin update immediately gets picked up by excavator, applied to people's repos
3. Jetbrains takes two business days to approve our intellij plugin.
4. Every single time Gradle is refreshed, people get the popup to update to the new IntelliJ version that _is not available yet_.

crogoz: that's a good point, I'll see if I can check that the version exists before opening the popup.
But that means we cannot really rely on the minVersion. Probably the best would be if the plugin will always maintain backwards compatibility with the intellij plugin and I can add a test for the logic [here](https://github.com/palantir/gradle-jdks/blob/develop/gradle-jdks/src/main/java/com/palantir/gradle/jdks/PalantirGradleJdksIdeaPlugin.java#L28) that we won't update the min-version of a plugin if that version doesn't already exist.

crogoz: added a test that checks that the min-version intellij plugin: https://github.com/palantir/gradle-jdks/pull/410/files#diff-1d01bc54a43b33238be402e57110068b45d10d5b2c7f88104b1c1e67080c1a53R36. This should stop us from bumping the required min-version in externalDependencies.xml file.

I could add the check (probably cached + expiration) [here](https://github.com/palantir/gradle-jdks/pull/410/files#diff-c40f3e7fde707d4f846f1cb40c95c063ac736bb45bc484fa9c9b712dfa537c5aR67) in case we are going to "recall"(delete) a version, but I am not sure if we need it atm.

PR #416, opened 2024-09-17
Intellij plugin doesn't need IDE restart anymore

crogoz:
## Before this PR

Any update of the palantir-gradle-jdk plugin needed a restart of Intellij. 

## After this PR
In preparation for https://github.com/palantir/gradle-jdks/pull/410 (prompting the user to update the version of the plugin), we are making the plugin dynamic (https://plugins.jetbrains.com/docs/intellij/dynamic-plugins.html) such that upgrading the plugin version won't need a restart of the IDE anymore. 

==COMMIT_MSG==
Intellij plugin doesn't need IDE restart anymore
==COMMIT_MSG==

## Possible downsides?
Not possible to test updating the plugin using `./gradlew runIde`. I ran the verification as per https://plugins.jetbrains.com/docs/intellij/dynamic-plugins.html#restrictions & ade sure we are cleaning up all the resources (https://plugins.jetbrains.com/docs/intellij/dynamic-plugins.html#resource-cleanup).

### DIFF
--- a/idea-plugin/src/main/java/com/palantir/gradle/jdks/GradleJdksProjectService.java
+++ b/idea-plugin/src/main/java/com/palantir/gradle/jdks/GradleJdksProjectService.java
@@ -29,13 +29,14 @@ import com.intellij.execution.ui.ConsoleViewContentType;
 import com.intellij.notification.Notification;
 import com.intellij.notification.NotificationGroupManager;
 import com.intellij.notification.NotificationType;
+import com.intellij.openapi.Disposable;
 import com.intellij.openapi.components.Service;
 import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.project.Project;
-import com.intellij.openapi.util.Disposer;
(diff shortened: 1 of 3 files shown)


idea-plugin/src/main/java/com/palantir/gradle/jdks/GradleJdksProjectService.java
```diff
@@ -111,6 +111,7 @@ public void maybeSetupGradleJdks() {
         TasksKt.withBackgroundProgress(
                 project,
                 "Gradle JDK Setup",
+                TaskCancellation.nonCancellable(),
```
crogoz, line 114: Previously the user had a button "x" to cancel the background task, but even if they clicked the button nothing would happen until the `./gradle-jdks-setup` script finished running.
Now, the button won't be shown anymore. 
Screenshot 2024-09-17 at 12 12 05 PM

CRogers (approved):

PR #420, opened 2024-09-23
Fixes caCerts retrieval in Windows

crogoz:
## Before this PR

In: https://github.com/palantir/gradle-jdks/commit/90cbeb28906c326a140f8a969ae721725e0945a6#diff-f0333751b6ca40fe3d118c9f2fce18e089d40d62e5bb0649a7a0635b82e5df56L158-R199 there was a regression in the way we would get the caCerts for Windows cases. We are using this function to configure the caCerts in: https://github.com/palantir/gradle-jdks/blob/1de223441106953fbee374810acceba8855c9735/gradle-jdks/src/main/java/com/palantir/gradle/jdks/PalantirCaPlugin.java#L50. 

## After this PR

==COMMIT_MSG==
Fixes caCerts retrieval in Windows
==COMMIT_MSG==


gradle-jdks-settings/build.gradle
```diff
@@ -55,5 +55,5 @@ publishing.publications {
 tasks.withType(Test) {
     environment("HOME", "/tmp/gradleJdksSettingsTest")
     systemProperty 'user.home', '/tmp/gradleJdksSettingsTest'
+    dependsOn project(':gradle-jdks').tasks.withType(PluginUnderTestMetadata.class)
```
crogoz, line 58: the tests in the settings plugin applies the plugins from gradle-jdks as well: https://github.com/palantir/gradle-jdks/blob/develop/gradle-jdks-settings/src/test/groovy/com/palantir/gradle/jdks/settings/ToolchainJdksSettingsPluginTest.groovy#L35, we we need to make sure the file `plugin-under-metadata.xml` file with the `implementation-classpath` exists

esword (approved):

PR #423, opened 2024-09-24
add support for OpenJDK Early Access builds

bjlaub:
## Before this PR
Only Zulu, Corretto, and Graal CE are supported

## After this PR
==COMMIT_MSG==
add support for OpenJDK Early Access builds
==COMMIT_MSG==

## Possible downsides?
Early Access builds should really only be used for experimentation at the moment. However it may be generally advantageous to have this pattern available for future EA builds from OpenJDK projects (e.g. Valhalla or any as-yet-unreleased OpenJDK build, which could be used for prototyping before features land in a full JDK release).


crogoz (dismissed):

bjlaub:
i think this might still need some tweaking to allow the base URL to be correctly overridden in certain scenarios

gradle-jdks-distributions/src/main/java/com/palantir/gradle/jdks/JdkDistributionName.java
```diff
@@ -24,7 +24,8 @@
 public enum JdkDistributionName {
     AZUL_ZULU,
     AMAZON_CORRETTO,
-    GRAALVM_CE;
+    GRAALVM_CE,
+    LOOM_EA;
```
CRogers: Rather than loom EA specifically, can we just have one for all OpenJDK EAs (and have loom in the version string?)

CRogers: Looks like it is in the version string? `24-loom+7-60`

bjlaub: Yeah I think I can make this work. The URLs are ever so slightly different for e.g. [JDK24 EA builds](https://download.java.net/java/early_access/jdk24/17/GPL/openjdk-24-ea+17_linux-aarch64_bin.tar.gz) versus [project-specific builds](https://download.java.net/java/early_access/loom/7/openjdk-24-loom+7-60_linux-aarch64_bin.tar.gz), but I think we can work around that.

bjlaub: @CRogers i think https://github.com/palantir/gradle-jdks/pull/423/commits/e0fb0db5a32b2bef1b6fc0091892bd2233fd4795 should do the trick

carterkozak (dismissed):
Approving with DNM to unblock a release candidate

carterkozak (dismissed):

PR #447, opened 2024-10-24
Set `JAVA_HOME` in gradle-jdks-setup script

crogoz:
## Before this PR

In case there is no JAVA_HOME, then ./gradlew will look at `java` executable (https://github.com/palantir/gradle-jdks/blob/develop/gradlew#L142). If no `java` is installed, then the ./gradlew command will fail.

example: https://app.circleci.com/pipelines/github/palantir/gradle-jdks/2274/workflows/66d6abd4-88f9-4525-aed9-e91363f1921a/jobs/6532 when trying to drop the "Install Java" step from the circle-ci template (https://github.com/palantir/gradle-jdks/commit/369952b98baaf28aad95c7b6c322e062f8567b2f#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47L93-L98)

## After this PR

Gradle JDK Automanagement will also set `JAVA_HOME` used by `./gradlew` to launch Gradle (https://github.com/palantir/gradle-jdks/blob/develop/gradlew#L128-L150).

==COMMIT_MSG==
gradle-jdks-setup.sh script sets `JAVA_HOME` to gradle JDK
==COMMIT_MSG==


gradle-jdks-setup/src/integrationTest/java/com/palantir/gradle/jdks/setup/GradleJdkInstallationSetupIntegrationTest.java
```diff
@@ -134,28 +132,14 @@ private Path setupGradleDirectoryStructure(Os os) throws IOException {
...
-        caResources
-                .readPalantirRootCaFromSystemTruststore()
-                .map(AliasContentCert::getContent)
-                .ifPresent(content -> {
-                    try {
-                        Files.write(workingDir.resolve("palantir.crt"), content.getBytes(StandardCharsets.UTF_8));
-                    } catch (IOException e) {
-                        throw new RuntimeException("Failed to write cert", e);
-                    }
-                });
```
CRogers, line 147: No longer need this as we just read certs from the system truststore right?

crogoz: yes

gradle-jdks-setup/src/integrationTest/resources/testing-script.sh
```diff
@@ -3,7 +3,18 @@
...
+if [ -f gradle/gradle-jdks-setup.sh ]; then
+    if ! . gradle/gradle-jdks-setup.sh; then
+        echo "Failed to set up JDK, running gradle/gradle-jdks-setup.sh failed with non-zero exit code"
+        exit 1
+    fi
+    # Setting JAVA_HOME to the gradle daemon to make sure gradlew uses this jdk for `JAVACMD`
+    JAVA_HOME="$GRADLE_DAEMON_JDK"
 fi
-echo "Example.com cert:" $(/root/.gradle/gradle-jdks/amazon-corretto-11.0.21.9.1/bin/keytool -keystore /root/.gradle/gradle-jdks/amazon-corretto-11.0.21.9.1/lib/security/cacerts -list -alias gradleJdks_example.com -storepass changeit)
+# <<< Gradle JDK setup <<<
```
CRogers: Is this manually placed here or is it kept up to date with `gradlew-patch.sh` somehow?

gradle-jdks-setup/build.gradle
```diff
@@ -42,7 +42,7 @@ tasks.integrationTest {
     dependsOn tasks.fatJar
 }
 
-tasks.build.dependsOn tasks.fatJar
+tasks.build.dependsOn tasks.fatJar, tasks.integrationTest
```
crogoz, line 45: the integration tests were not running before :(

gradle-jdks-setup/src/main/resources/gradle-jdks-setup.sh
```diff
@@ -74,6 +74,7 @@ gradle_daemon_jdk_distribution_local_path=$(read_value "$APP_GRADLE_DIR"/jdks/"$
 "$GRADLE_JDKS_HOME"/"$gradle_daemon_jdk_distribution_local_path"/bin/java -cp "$APP_GRADLE_DIR"/gradle-jdks-setup.jar com.palantir.gradle.jdks.setup.GradleJdkInstallationSetup daemonSetup "$APP_HOME" "$GRADLE_JDKS_HOME/$gradle_daemon_jdk_distribution_local_path"
 
 # [Used by ./gradlew only] Setting the Gradle Daemon Java Home to the JDK distribution
-set -- "-Dorg.gradle.java.home=$GRADLE_JDKS_HOME/$gradle_daemon_jdk_distribution_local_path" "$@"
+export GRADLE_DAEMON_JDK="$GRADLE_JDKS_HOME/$gradle_daemon_jdk_distribution_local_path"
```
crogoz, line 77: I am not setting the JAVA_HOME directly here as this script is called by the palantir-gradle-jdks intellij plugin as well and I don't think we should modify the JAVA_HOME there (when gradle jdk automanagement is set, then we also set the `Project SDK`)

CRogers: I am I right in thinking setting JAVA_HOME wouldn't really do anything bad though? We'd set the `JAVA_HOME` then the script would end to no real effect (when called from intellij in the before gradle step right?)

CRogers: Not that I think it really matters, the `export GRADLE_DAEMON_JDK` approach here is also fine.

CRogers (approved):

gradle-jdks-setup/src/main/resources/gradle-jdks-setup.sh
crogoz: > I am I right in thinking setting JAVA_HOME wouldn't really do anything bad though? We'd set the JAVA_HOME then the script would end to no real effect (when called from intellij in the before gradle step right?

I think you are right (I tested this initially inside Intellij) and it seemed fine. But I wouldn't want to add to any confusion if in the end something inside Intellij changes and the java home is somehow used.

PR #452, opened 2024-10-30
[Fix] JDK is successfully installed if `bin/java` files exist

crogoz:
## Before this PR

Context: https://pl.ntr/2p9
We consider that a JDK was successfully installed if the directory exists. 
However, this is not always true. eg. if an exception happens while the JDK is copied over to the destination directory (https://github.com/palantir/gradle-jdks/blob/develop/gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/GradleJdkInstallationSetup.java#L130). The JDK directory might not contain the `bin/java` executable. 

## After this PR

A JDK directory is considered successfully installed if `bin/java` exists . 
I am also doing a copy to a temporary directory and an atomic move to the destination directory.
A `sentinel` file is added at the end of a successful installation - I think we might want to use this in the future to check if the jdk was successfully installed or not. 

Note: I am not considering for now only the `sentinel` file to be the source of truth because of backwards compatibility (the jdks we installed so far won't have the `sentinel` file).

==COMMIT_MSG==
[Fix] JDK is successfully installed if `bin/java` exists
==COMMIT_MSG==

### DIFF
--- a/gradle-jdks-setup/src/integrationTest/java/com/palantir/gradle/jdks/setup/GradleJdkInstallationSetupIntegrationTest.java
+++ b/gradle-jdks-setup/src/integrationTest/java/com/palantir/gradle/jdks/setup/GradleJdkInstallationSetupIntegrationTest.java
@@ -19,7 +19,6 @@ package com.palantir.gradle.jdks.setup;
 import static org.assertj.core.api.Assertions.assertThat;
 
 import com.google.common.base.Splitter;
-import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
 import com.palantir.gradle.jdks.AmazonCorrettoJdkDistribution;
 import com.palantir.gradle.jdks.JdkPath;
@@ -44,7 +43,6 @@ public class GradleJdkInstallationSetupIntegrationTest {
 
     private static final String JDK_VERSION = "11.0.21.9.1";
     private static final Arch ARCH = CurrentArch.get();
-    private static final Optional<Path> NO_MOUNT = Optional.empty();
     private static final String CORRETTO_DISTRIBUTION_URL_ENV = "CORRETTO_DISTRIBUTION_URL";
     private static final AmazonCorrettoJdkDistribution CORRETTO_JDK_DISTRIBUTION = new AmazonCorrettoJdkDistribution();
     private static final boolean DO_NOT_INSTALL_CURL = false;
@@ -56,26 +54,25 @@ public class GradleJdkInstallationSetupIntegrationTest {
     @Test
     public void can_setup_jdks_centos() throws IOException, InterruptedException {
         setupGradleDirectoryStructure(Os.LINUX_GLIBC);
-        dockerBuildAndRunTestingScript("centos:7", "/bin/bash", DO_NOT_INSTALL_CURL, NO_MOUNT);
+        dockerBuildAndRunTestingScript("centos:7", "/bin/bash", DO_NOT_INSTALL_CURL, false);
     }
 
     @Test
     public void can_setup_jdks_ubuntu() throws IOException, InterruptedException {
         setupGradleDirectoryStructure(Os.LINUX_GLIBC);
-        dockerBuildAndRunTestingScript("ubuntu:20.04", "/bin/bash", INSTALL_CURL, NO_MOUNT);
+        dockerBuildAndRunTestingScript("ubuntu:20.04", "/bin/bash", INSTALL_CURL, false);
     }
 
     @Test
     public void can_reinstall_jdks_ubuntu() throws IOException, InterruptedException {
         setupGradleDirectoryStructure(Os.LINUX_GLIBC);
-        Path dir = Files.createDirectories(workingDir.resolve("amazon-corretto-11.0.21.9.1"));
-        dockerBuildAndRunTestingScript("ubuntu:20.04", "/bin/bash", INSTALL_CURL, Optional.of(dir));
+        dockerBuildAndRunTestingScript("ubuntu:20.04", "/bin/bash", INSTALL_CURL, true);
     }
 
     @Test
     public void can_setup_jdks_alpine() throws IOException, InterruptedException {
         setupGradleDirectoryStructure(Os.LINUX_MUSL);
-        dockerBuildAndRunTestingScript("alpine:3.16.0", "/bin/sh", DO_NOT_INSTALL_CURL, NO_MOUNT);
+        dockerBuildAndRunTestingScript("alpine:3.16.0", "/bin/sh", DO_NOT_INSTALL_CURL, false);
     }
 
     private Path setupGradleDirectoryStructure(Os os) throws IOException {
@@ -163,7 +160,7 @@ public class GradleJdkInstallationSetupIntegrationTest {
     }
 
     private void dockerBuildAndRunTestingScript(
-            String baseImage, String shell, boolean installCurl, Optional<Path> mountGradleJdkDir)
+            String baseImage, String shell, boolean installCurl, boolean addEmptyGradleJdkDir)
             throws IOException, InterruptedException {
         Path dockerFile = Path.of("src/integrationTest/resources/template.Dockerfile");
         String dockerImage = String.format("jdk-test-%s", baseImage);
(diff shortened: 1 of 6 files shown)


gradle-jdks-setup/src/main/resources/gradle-jdks-functions.sh
```diff
@@ -123,44 +123,49 @@ install_and_setup_jdks() {
...
-      # Finding the java_home
-      java_home=$(get_java_home "$in_progress_dir")
-      "$java_home"/bin/java -cp "$scripts_dir"/gradle-jdks-setup.jar com.palantir.gradle.jdks.setup.GradleJdkInstallationSetup jdkSetup "$jdk_installation_directory" || die "Failed to set up JDK $jdk_installation_directory"
-      echo "Successfully installed JDK distribution in $jdk_installation_directory"
+    elif [ ! -f "$jdk_installation_directory/bin/java" ]; then
+      echo "Java executable not found in $jdk_installation_directory/bin/java, re-installing the JDK...."
+    else
+      echo "JDK installation $jdk_installation_directory was already configured"
+      continue
+    fi
```
crogoz, line 132: the github diff is weird, the only changes I have made is adding the if-else conditions to check if the directory and/or bin/java  files exists at the top. the rest of the code is unchanged

gradle-jdks-setup/src/integrationTest/java/com/palantir/gradle/jdks/setup/GradleJdkInstallationSetupIntegrationTest.java
```diff
@@ -153,7 +159,8 @@ private static void writeFileContent(Path path, String content) throws IOExcepti
         Files.writeString(path, content + "\n");
     }
 
-    private void dockerBuildAndRunTestingScript(String baseImage, String shell, boolean installCurl)
+    private void dockerBuildAndRunTestingScript(
+            String baseImage, String shell, boolean installCurl, boolean addEmptyGradleJdkDir)
```
CRogers: I assumed this meant the `~/.gradle/gradle-jdks` dir - maybe call this just `addJdkDir`?

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/GradleJdkInstallationSetup.java
```diff
@@ -121,16 +123,39 @@ private static boolean copy(ILogger logger, Path destinationJdkInstallationDirec
...
+                Files.move(
+                        tempDirWithPrefix,
+                        destinationJdkInstallationDirectory,
+                        StandardCopyOption.REPLACE_EXISTING,
+                        StandardCopyOption.ATOMIC_MOVE);
+            } catch (AtomicMoveNotSupportedException e) {
+                logger.log("Cannot move the directory to the final directory, attempting to do a non-atomic operation");
+                Files.move(tempDirWithPrefix, destinationJdkInstallationDirectory, StandardCopyOption.REPLACE_EXISTING);
+            }
+            createSentinelFile(destinationJdkInstallationDirectory);
```
CRogers: What is this sentinel file used for?

CRogers: I don't think it's used elsewhere in this diff and now we have the atomic move the directory/java binary existing should be enough.

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/GradleJdkInstallationSetup.java
```diff
@@ -121,16 +123,39 @@ private static boolean copy(ILogger logger, Path destinationJdkInstallationDirec
...
-            if (Files.exists(destinationJdkInstallationDirectory)) {
+            if (Files.exists(destinationJdkInstallationDirectory)
+                    && Files.exists(destinationJdkInstallationDirectory.resolve("bin/java"))) {
                 logger.log(String.format("Distribution URL %s already exists", destinationJdkInstallationDirectory));
                 return false;
             }
             logger.log(
                     String.format("Copying JDK from %s into %s", currentJavaHome, destinationJdkInstallationDirectory));
-            FileUtils.copyDirectory(currentJavaHome, destinationJdkInstallationDirectory);
+            Path tempDirWithPrefix = Files.createTempDirectory("gradle-jdks");
```
CRogers: Atomic moves generally only work when the source and destination for the filesystem object are on the same filesystem. If they're not it's usually actually a copy. I think here this risks copying to a tmp filesystem, then copying again to the filesystem `~/.gradle/gradle-jdks` is mounted on rather than moving to the final destination.

To work around this, I'd make a `tmp` dir at `~/.gradle/gradle-jdks/tmp` and copy it to a subdir in there first before moving. It should definitely be on the same filesystem then.

CRogers: I would also delete the temp dir if it still exists in a try-finally to make sure it's cleaned even if there's an exception. Don't want an ever increasing number of these.

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/FileUtils.java
```diff
@@ -65,13 +66,13 @@ public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes _attrs) t
 
             @Override
             public FileVisitResult visitFile(Path file, BasicFileAttributes _attrs) throws IOException {
-                Files.copy(file, destination.resolve(source.relativize(file)));
+                Files.copy(file, destination.resolve(source.relativize(file)), StandardCopyOption.REPLACE_EXISTING);
```
CRogers, line 69: Do we still need to replace-existing with the copy to a temp location and then move?

gradle-jdks-setup/src/main/resources/gradle-jdks-functions.sh
```diff
@@ -123,44 +123,49 @@ install_and_setup_jdks() {
...
-        die "ERROR: Neither curl nor wget are installed, Could not set up JAVA_HOME"
-      fi
-      cd - || exit
-
-      # Finding the java_home
-      java_home=$(get_java_home "$in_progress_dir")
-      "$java_home"/bin/java -cp "$scripts_dir"/gradle-jdks-setup.jar com.palantir.gradle.jdks.setup.GradleJdkInstallationSetup jdkSetup "$jdk_installation_directory" || die "Failed to set up JDK $jdk_installation_directory"
-      echo "Successfully installed JDK distribution in $jdk_installation_directory"
+    elif [ ! -f "$jdk_installation_directory/bin/java" ]; then
+      echo "Java executable not found in $jdk_installation_directory/bin/java, re-installing the JDK...."
```
CRogers, line 128: I think worth having a comment about the possibly broken existing JDKs.

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/GradleJdkInstallationSetup.java
crogoz: I am thinking that the sentinel file is the flag that tells us if the installation finished successfully. I was thinking that in the future (if we add the check now , we would re-install jdks that are already correctly installed on user's laptops) we might assume that the installation was successful if the file was created.

CRogers: I would hope that since we move the JDK now it will definitely be fine as the move will mean the directory is either there and correct or not there, rather than just partially copied.

crogoz: ack, I'll remove this then

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/FileUtils.java
crogoz: I think it is still better to keep the replace-existing, otherwise if some crash leaves the tmp location un-deleted, then we can recover from it.

gradle-jdks-setup/src/main/resources/gradle-jdks-functions.sh
crogoz: this file will be added to all the repos (it is used both by the docker images and the gradle jdk enabled repos). Is it OK to mention the possibly broken JDKs in a comment ?

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/GradleJdkInstallationSetup.java
```diff
@@ -121,16 +122,37 @@ private static boolean copy(ILogger logger, Path destinationJdkInstallationDirec
...
+            // 1. Copying the currentJavaHome to a temporary directory inside the jdksInstallationDirectory
+            // 2. Atomic move of the temporary directory to the destination directory
+            try {
+                Files.createDirectories(tempCopyDir);
+                FileUtils.copyDirectory(currentJavaHome, tempCopyDir);
+                Files.move(
+                        tempCopyDir,
+                        destinationJdkInstallationDirectory,
+                        StandardCopyOption.REPLACE_EXISTING,
+                        StandardCopyOption.ATOMIC_MOVE);
```
CRogers, line 147: The common filesystems (ieg APFS/NTFS/ext4) all support atomic moves... but perhaps we should handle the `AtomicMoveNotSupportedException` like before just incase this ends up on a different filesystem that doesn't?

CRogers (approved):

PR #453, opened 2024-10-31
Rename check & setupJdks to disambiguate task shortcuts 

crogoz:
## Before this PR

Context: https://pl.ntr/2pc
Users were using `gw cJ` to run the compileJava task. `checkJava` is now matching this shortcut as well.

## After this PR

Change task names to disambiguate used shortcuts 
==COMMIT_MSG==
Rename check & setupJdks to disambiguate task shortcuts 
==COMMIT_MSG==

### DIFF
--- a/gradle-jdks/src/main/java/com/palantir/gradle/jdks/ToolchainsPlugin.java
+++ b/gradle-jdks/src/main/java/com/palantir/gradle/jdks/ToolchainsPlugin.java
@@ -116,7 +116,7 @@ public final class ToolchainsPlugin implements Plugin<Project> {
             task.finalizedBy(wrapperPatcherTask);
         });
 
-        TaskProvider<Task> checkJdksLifecycle = rootProject.getTasks().register("checkJdks", Task.class, task -> {
+        TaskProvider<Task> checkJdksLifecycle = rootProject.getTasks().register("checkGradleJdks", Task.class, task -> {
             task.setDescription("Lifecycle task that checks the Gradle JDK configurations.");
             task.setGroup(GRADLE_JDK_GROUP);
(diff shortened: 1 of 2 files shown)


gradle-jdks/src/main/java/com/palantir/gradle/jdks/ToolchainsPlugin.java
```diff
@@ -127,7 +127,7 @@ public void apply(Project rootProject) {
                 .named(LifecycleBasePlugin.CHECK_TASK_NAME)
                 .configure(check -> check.dependsOn(checkJdksLifecycle));
 
-        rootProject.getTasks().register("setupJdks", SetupJdksTask.class, setupJdksTask -> {
+        rootProject.getTasks().register("setupGradleJdks", SetupJdksTask.class, setupJdksTask -> {
```
crogoz: changed the name for consistency

felixdesouza: do you have to change things in the intellij plugin?

CRogers: Isn't this going to break excavators?

crogoz: > Isn't this going to break excavators?

you are right, thanks for catching this! I reverted this rename. 

> do you have to change things in the intellij plugin?

the intellij plugin is calling directly the `gradle-jdks-setup.sh` script

CRogers (approved):

PR #458, opened 2024-11-05
Remove noisy output

pkoenig10:
This output is emitted on every Gradle build for every JDK.

```
JDK installation /Users/pkoenig/.gradle/gradle-jdks/amazon-corretto-11.0.25.9.1 was already configured
JDK installation /Users/pkoenig/.gradle/gradle-jdks/amazon-corretto-17.0.13.11.1 was already configured
JDK installation /Users/pkoenig/.gradle/gradle-jdks/amazon-corretto-21.0.5.11.1 was already configured
```

### DIFF
--- a/gradle/gradle-jdks-functions.sh
+++ b/gradle/gradle-jdks-functions.sh
@@ -127,7 +127,6 @@ install_and_setup_jdks() {
     elif [ ! -f "$jdk_installation_directory/bin/java" ]; then
       echo "Java executable not found in $jdk_installation_directory/bin/java, re-installing the JDK...."
     else
-      echo "JDK installation $jdk_installation_directory was already configured"
       continue
     fi
     # Download and extract the distribution into a temporary directory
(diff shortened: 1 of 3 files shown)


gradle/gradle-jdks-functions.sh
```diff
@@ -127,7 +127,6 @@ install_and_setup_jdks() {
     elif [ ! -f "$jdk_installation_directory/bin/java" ]; then
       echo "Java executable not found in $jdk_installation_directory/bin/java, re-installing the JDK...."
     else
-      echo "JDK installation $jdk_installation_directory was already configured"
```
CRogers: Mmm, this is noisy, it was added in #452. We should probably have some test that we don't print anything when the JDK is already installed (to be added later).

CRogers:
:+1:

CRogers:
:+1:

felixdesouza:
👍

PR #464, opened 2024-11-13
Basic check for insecure curl/wget requests

crogoz:
## After this PR

Extra checks that would avoid bugs like here: https://github.com/palantir/gradle-jdks/pull/452/files#diff-eab8ad98fab9b4edb9ec2349b5d7a41d5c93bf016e9f409511c6ef497d8c2facR146 where insecure flags were introduced in curl requests

==COMMIT_MSG==
Basic check for insecure curl/wget requests
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-464.v2.yml
+++ b/changelog/@unreleased/pr-464.v2.yml
@@ -1,5 +1,0 @@
-type: fix
-fix:
-  description: Basic check for insecure curl/wget requests
-  links:
-  - https://github.com/palantir/gradle-jdks/pull/464
--- a/changelog/@unreleased/pr-464.v2.yml
+++ b/changelog/@unreleased/pr-464.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Basic check for insecure curl/wget requests
+  links:
+  - https://github.com/palantir/gradle-jdks/pull/464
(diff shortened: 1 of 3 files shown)


gradle-jdks-setup/src/test/java/com/palantir/gradle/jdks/setup/GradleJdksVulnerabilityChecks.java
```diff
@@ -0,0 +1,42 @@
...
+import org.junit.jupiter.api.Test;
+
+public class GradleJdksVulnerabilityChecks {
+    private static final Pattern CURL_INSECURE_PATTERN = Pattern.compile("curl\\s.*(-k|--insecure)");
+    private static final Pattern WGET_INSECURE_PATTERN = Pattern.compile("wget\\s.*--no-check-certificate");
+
+    @Test
+    public void calls_do_not_use_insecure_connections() throws IOException {
+        List<String> lines = Files.readAllLines(Path.of("src/main/resources/gradle-jdks-functions.sh"));
+        for (String line : lines) {
```
CRogers: I think we can just `File.readString` here and match on the whole string?

changelog/@unreleased/pr-464.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: fix
+fix:
+  description: Basic check for insecure curl/wget requests
```
CRogers: Probably don't need changelog or a release for this.

PR #465, opened 2024-11-13
Preserve empty line in patched file

crogoz:
## Before this PR

Gradlew patch file doesn't contain the empty line at the end of the file. example: https://github.com/palantir/gradle-jdks/pull/411/files#diff-e9721dc750619a21053ddea8a5d04929a608877d8c5daec1b57d243d3424e745R253

## After this PR

Empty line at the end of a patched file.
==COMMIT_MSG==
Preserve empty line in patched file
==COMMIT_MSG==


gradle-jdks-setup-common/src/main/java/com/palantir/gradle/jdks/setup/common/GradleJdksPatchHelper.java
```diff
@@ -69,30 +68,23 @@ public static List<String> getLinesWithoutPatch(List<String> initialLines) {
...
-            List<String> initialLines = Files.readAllLines(filePath);
-            List<String> linesWithoutPatch = getLinesWithoutPatch(initialLines);
-            if (!linesWithoutPatch.equals(initialLines)) {
-                Files.write(
-                        filePath,
-                        String.join("\n", linesWithoutPatch).getBytes(StandardCharsets.UTF_8),
-                        StandardOpenOption.TRUNCATE_EXISTING,
-                        StandardOpenOption.WRITE);
-            }
+            Files.write(
```
CRogers: `writeString` and you can avoid the charset/bytes.

gradle-jdks-setup-common/src/main/java/com/palantir/gradle/jdks/setup/common/GradleJdksPatchHelper.java
```diff
@@ -69,30 +68,23 @@ public static List<String> getLinesWithoutPatch(List<String> initialLines) {
...
-    public static byte[] getContentWithPatch(List<String> initialLines, List<String> patchLines, int insertIndex) {
+    private static String getContentWithPatch(List<String> initialLines, List<String> patchLines, int insertIndex) {
         List<String> newLines = new ArrayList<>(initialLines.size() + patchLines.size());
         newLines.addAll(initialLines.subList(0, insertIndex));
         newLines.addAll(patchLines);
         newLines.addAll(initialLines.subList(insertIndex, initialLines.size()));
-        return newLines.stream()
-                .collect(Collectors.joining(System.lineSeparator()))
-                .getBytes(StandardCharsets.UTF_8);
+        return newLines.stream().collect(Collectors.joining(System.lineSeparator())) + "\n";
```
CRogers: I think we need to check whether the existing file had a newline or not? Unless we're just depending on `./gradlew` always having a newline.

crogoz: addressed in [e268e40](https://github.com/palantir/gradle-jdks/pull/465/commits/e268e40a0a07ed44914204e08e561467ba2f1974), but I think in general it is nice for files to end with a newline and I think it is safe to assume that `./gradlew` will always have a new line

CRogers (approved):

PR #472, opened 2024-11-22
Add `jdkMajorVersionsToUse` and `daemonJdkOnly` for gradle jdk directory generation

crogoz:
## Before this PR

In cases where we only need to use the jdk for the daemon, we are generating all the jdk directories in `gradle/jdks`. However not all of them are needed. Moreover, we shouldn't install all the jdks that are present in the `latest-gradle-jdks` extension, we should instead only install the used jdks (eg. for the compilation, testing etc.)

## After this PR

- Added a `jdkMajorVersionsToUse ` set property with all the JavaLanagugeVersions for which we should generate the gradle configuration files. Based on these configuration files, we will install & configure the jdks. Used in the future to only include the versions set up by `baseline-java-versions`
- Add a `daemonJdkOnly()` method to the `JdksExtension` that will only generate the jdks required for the gradle daemon jdk version. Context: https://pl.ntr/2pB

==COMMIT_MSG==
Add `jdkMajorVersionsToUse` and `daemonJdkOnly` for gradle jdk directory generation
==COMMIT_MSG==

### DIFF
--- a/gradle-jdks-setup/README.md
+++ b/gradle-jdks-setup/README.md
@@ -89,8 +89,8 @@ jdks {
 ```
 jdks {
     daemonTarget = <version eg. 11/17/21>
-    daemonOnly() # optional, only installs & configures the daemon JDK version
-    jdkVersionsToUse = <set of versions eg. 11/17/21> # optional, the set of java versions to install, by default it will contain all the major versions configured in step 3.
+    daemonJdkOnly() # optional, only installs & configures the daemon JDK version
+    jdkVersionsToUse = [11, 17, 21, 23] # optional, the set of java versions to install, by default it will contain all the major versions configured in step 3.
 }
 ```
 5. Enable the jdk setup by adding the following to `gradle.properties`:
--- a/gradle-jdks-setup/README.md
+++ b/gradle-jdks-setup/README.md
@@ -90,7 +90,7 @@ jdks {
 jdks {
     daemonTarget = <version eg. 11/17/21>
     daemonJdkOnly() # optional, only installs & configures the daemon JDK version
-    jdkVersionsToUse = <set of versions eg. 11/17/21> # optional, the set of java versions to install, by default it will contain all the major versions configured in step 3.
+    jdkVersionsToUse = [11, 17, 21, 23] # optional, the set of java versions to install, by default it will contain all the major versions configured in step 3.
 }
 ```
 5. Enable the jdk setup by adding the following to `gradle.properties`:
--- a/gradle-jdks-setup/README.md
+++ b/gradle-jdks-setup/README.md
@@ -89,7 +89,7 @@ jdks {
 ```
 jdks {
     daemonTarget = <version eg. 11/17/21>
-    daemonOnly() # optional, only installs & configures the daemon JDK version
+    daemonJdkOnly() # optional, only installs & configures the daemon JDK version
     jdkVersionsToUse = <set of versions eg. 11/17/21> # optional, the set of java versions to install, by default it will contain all the major versions configured in step 3.
 }
 ```
--- a/gradle-jdks-setup/README.md
+++ b/gradle-jdks-setup/README.md
@@ -89,6 +89,8 @@ jdks {
 ```
 jdks {
     daemonTarget = <version eg. 11/17/21>
+    daemonOnly() # optional, only installs & configures the daemon JDK version
+    jdkVersionsToUse = <set of versions eg. 11/17/21> # optional, the set of java versions to install, by default it will contain all the major versions configured in step 3.
 }
 ```
 5. Enable the jdk setup by adding the following to `gradle.properties`:
--- a/gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdksExtension.java
+++ b/gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdksExtension.java
@@ -36,6 +36,7 @@ public abstract class JdksExtension {
     private final MapProperty<String, String> caCerts;
     private final DirectoryProperty jdkStorageLocation;
     private final Property<JavaLanguageVersion> daemonTarget;
+    private final Property<Boolean> daemonJdkOnly;
 
     @Inject
     protected abstract ObjectFactory getObjectFactory();
@@ -57,9 +58,11 @@ public abstract class JdksExtension {
         this.jdkStorageLocation = SynchronizedInterface.synchronizeAllInterfaceMethods(
                 DirectoryProperty.class, getObjectFactory().directoryProperty());
         this.daemonTarget = getObjectFactory().property(JavaLanguageVersion.class);
+        this.daemonJdkOnly = getObjectFactory().property(Boolean.class);
         this.getCaCerts().finalizeValueOnRead();
         this.getJdkStorageLocation().finalizeValueOnRead();
         this.getDaemonTarget().finalizeValueOnRead();
+        this.getDaemonJdkOnly().finalizeValueOnRead();
     }
 
     public final Property<JavaLanguageVersion> getDaemonTarget() {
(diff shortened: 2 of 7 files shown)


gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdksExtension.java
```diff
@@ -74,6 +77,14 @@ public final void setDaemonTarget(int value) {
         getDaemonTarget().set(JavaLanguageVersion.of(value));
     }
 
+    public final Property<Boolean> getDaemonJdkOnly() {
+        return daemonJdkOnly;
+    }
+
+    public final void setDaemonJdkOnly(Boolean value) {
+        getDaemonJdkOnly().set(value);
+    }
```
CRogers: I'd actually hide this property maybe so it doesn't become a breaking change if we do further baseline-java-versions integration we talked about. ie no readme for it. No setter, we can make a method called `daemonJdkOnly()` that sets it to true. Then later we can make this method do `jdkMajorVersionsToUse.set(daemonTargert.map(Set::of))` later and no breaks.

CRogers: Actually, we could have this `jdkMajorVersionsToUse` property right now, it moves the logic from `JdkConfigurationDistributor`. Later we just give it a different default value (now it just needs to get all the keys from `jdks` as per in `JdkConfigurationDistributor`).

CRogers: I think `jdkMajorVersionsToUse` is a generally useful thing that could be used by external people (probably even worth documenting that and `daemonJdkOnly`).

gradle-jdks-setup/README.md
```diff
@@ -89,6 +89,8 @@ jdks {
 ```
 jdks {
     daemonTarget = <version eg. 11/17/21>
+    daemonOnly() # optional, only installs & configures the daemon JDK version
```
felixdesouza:
```suggestion
    daemonJdkOnly() # optional, only installs & configures the daemon JDK version
```

gradle-jdks-setup/README.md
```diff
@@ -89,6 +89,8 @@ jdks {
...
+    jdkVersionsToUse = <set of versions eg. 11/17/21> # optional, the set of java versions to install, by default it will contain all the major versions configured in step 3.
```
felixdesouza:
```suggestion
    jdkVersionsToUse = [11, 17, 21, 23] # optional, the set of java versions to install, by default it will contain all the major versions configured in step 3.
```

largely just making sure that you can copy paste the example without syntax errors

gradle-jdks/src/test/groovy/com/palantir/gradle/jdks/GradleJdkToolchainsIntegrationTest.groovy
```diff
@@ -156,6 +158,9 @@ class GradleJdkToolchainsIntegrationTest extends GradleJdkIntegrationSpec {
         Path daemonJvm = workingDir().resolve("gradle-jdks").resolve(daemonJdkFileName).toAbsolutePath()
         gradleHomeOutput.contains("java.home: ${daemonJvm}")
 
+        then: 'generates directories for all jdk versions'
+        Files.list(projectDir.toPath().resolve("gradle/jdks")).map { it -> it.getFileName().toString() }.collect(Collectors.toList()).containsAll(List.of("11", "17", "21"))
```
felixdesouza: surely spotless kills you here

felixdesouza: oh nvm, it's groovy

felixdesouza: can we just put the `.map`, `.collect` and `.containsAll` on separate lines pls 🙏

gradle-jdks/src/test/groovy/com/palantir/gradle/jdks/GradleJdkToolchainsIntegrationTest.groovy
```diff
@@ -175,6 +180,33 @@ class GradleJdkToolchainsIntegrationTest extends GradleJdkIntegrationSpec {
         gradleVersionNumber << GRADLE_TEST_VERSIONS
     }
 
+
+    def '#gradleVersionNumber: only generates daemon jdk'() {
+        gradleVersion = gradleVersionNumber
+        setupJdksHardcodedVersions('11')
```
felixdesouza, line 189: is this just setting up the directories? or does it do something else? it's hard to tell that this isn't the default behaviour if you didn't add `daemonJdkOnly` ?

gradle-jdks/src/test/groovy/com/palantir/gradle/jdks/GradleJdkToolchainsIntegrationTest.groovy
```diff
@@ -175,6 +180,33 @@ class GradleJdkToolchainsIntegrationTest extends GradleJdkIntegrationSpec {
...
+        """.stripIndent(true)
+
+        file('gradle.properties') << 'palantir.jdk.setup.enabled=true'
+        file('src/main/java/Main.java') << java17PreviewCode
+
+        when:
+        runTasksSuccessfully('wrapper')
+
+        then: 'only jdk 11 is generated'
+        Files.list(projectDir.toPath().resolve("gradle/jdks")).allMatch { it -> it.endsWith("gradle/jdks/11")}
```
felixdesouza: nit: same thing re new lines

gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdksExtension.java
```diff
@@ -57,6 +63,10 @@ public JdksExtension() {
         this.jdkStorageLocation = SynchronizedInterface.synchronizeAllInterfaceMethods(
                 DirectoryProperty.class, getObjectFactory().directoryProperty());
         this.daemonTarget = getObjectFactory().property(JavaLanguageVersion.class);
+        this.jdkMajorVersionsToUse = getObjectFactory().setProperty(JavaLanguageVersion.class);
+        this.jdkMajorVersionsToUse.convention(Arrays.stream(JavaVersion.values())
+                .map(javaVersion -> JavaLanguageVersion.of(javaVersion.getMajorVersion()))
+                .collect(Collectors.toSet()));
```
felixdesouza, line 69: if I'm understanding correctly, this makes every jdk version available to use? missing a bit of context here, appreciate it's also just moved from `JdkDistributionConfigurator` so probably nothing actionable just a bit confused

crogoz: atm, we are generating files for all the jdks that are configured in the jdks extension (11, 17, 21, ...) regardless if they are used or not. When `./gradlew` script runs, it will install all the jdks for which we generated files. 

In the next PR (wip PR), I am going to set `jdkMajorVersionsToUse ` to the jdk versions we actually use (based on `baseline-java-versions` configurations)

gradle-jdks/src/test/groovy/com/palantir/gradle/jdks/GradleJdkToolchainsIntegrationTest.groovy
crogoz: without `daemonJdkOnly` (default behaviour) we are generating directories for all jdks (see previous test: https://github.com/palantir/gradle-jdks/pull/472/files#diff-d8cb60820e935e74203f71da27d54ca1e390a3b5cb379b2a15f807309a3a4018R162).

gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdksExtension.java
felixdesouza: ahh okay

felixdesouza (approved):

PR #479, opened 2024-12-09
[Gradle JDK Automanagement] Install only used JDKs

crogoz:
## Before this PR

Atm we are generating jdk configurations for all jdks that are [configured by gradle-jdks-latest](https://github.com/palantir/gradle-jdks-latest/blob/develop/gradle-jdks-latest/src/main/resources/latestjdks/latest-gradle-jdks.json). This means extra work if the repository only uses a subset of those JDKs. This is especially relevant in the context of circle VMs - we end up installing jdks that might not be used during circle ci jobs.

## After this PR

Continuation for: https://github.com/palantir/gradle-jdks/pull/472. 
- We are only generating the jdk configuration files for the used java versions - depends on the application of the `gradle-baseline` plugins. 
- Adding an command line option `gw generateGradleJdkConfigs  --includeAllJdks ` that can generate all the jdk configuration files. Required by the excavator runs internally,
- Adding an command line option `gw generateGradleJdkConfigs  --includeVersion=<newJavaMajorVersion> ` that can generate the jdk configuration files for new major versions.

==COMMIT_MSG==
Install only used JDKs
==COMMIT_MSG==


gradle-jdks/src/main/java/com/palantir/gradle/jdks/GradleJdksConfigs.java
```diff
@@ -43,6 +45,15 @@ public abstract class GradleJdksConfigs extends DefaultTask {
...
     public static final String GRADLE_JDKS_FUNCTIONS_SCRIPT = "gradle-jdks-functions.sh";
 
+    /**
+     * Option to include a specific java major version when generating/checking the gradle jdk configuration files.
+     */
+    @Input
+    @Option(
+            option = "includeJava",
+            description = "Generates the jdk configuration directories for the Java major version.")
+    public abstract SetProperty<String> getIncludedJavaMajorVersion();
```
CRogers: I think we could avoid this if we were to regenerate the jdk configs at the same time (ie same gradle invocation in excavator) as bumping the versions? The problem should only present itself when changing the versions used, then trying to run gradle again but the wrong JDKs are included. I think we might be able to do that.

If that doesn't work, I think I'd prefer an option to just use all the configured versions rather than including specific ones. That way we don't need to keep updating excavators with specific versions, they can just include everything they need.

crogoz: > I think we could avoid this if we were to regenerate the jdk configs at the same time (ie same gradle invocation in excavator) as bumping the versions? The problem should only present itself when changing the versions used, then trying to run gradle again but the wrong JDKs are included. I think we might be able to do that.

I am not sure how that would work. If we want to bump a version of jdk used (eg. javaVersions `libraryTarget = 21`), when running any gradle task `./gradlew setupJdks` we don't have the jdk files configured so we are not installing anything. Then if jdk 21 doesn't exist, gradle will complain during the configuration phase that no toolchain was found for java == 21 when it was configuring eg. javaCompiler.

CRogers: Ah, makes sense, of course we have to do it in two invocations.

I think I'd just prefer the option then to use all the JDKs rather than opting into specific ones.

CRogers: I wonder if this should just be something you can set via an environment variable instead, so we can just set `export GRADLE_JDKS_INSTALL_ALL_JDKS=true` in the excavator and not have to worry about it setting it on all the different commands.

CRogers: Also probably need a test for this behaviour, whatever it ends up being.

crogoz:
waiting on https://github.com/palantir/gradle-jdks/pull/493 first & on updating the excavators.

gradle-jdks/src/main/java/com/palantir/gradle/jdks/ToolchainsPlugin.java
```diff
@@ -47,6 +53,9 @@ public void apply(Project rootProject) {
                             + "Gradle version in order to use the JDK setup.",
                     GradleJdksEnablement.MINIMUM_SUPPORTED_GRADLE_VERSION));
         }
+        if (areFlowActionsSupported()) {
+            rootProject.getPluginManager().apply(ToolchainFailureFlowActionsPlugin.class);
```
crogoz, line 57: the error message will only be available for repos >= 8.6

gradle-jdks/src/main/java/com/palantir/gradle/jdks/flow/ToolchainFlowAction.java
```diff
@@ -0,0 +1,92 @@
...
+                    })
+                    .filter(Optional::isPresent)
+                    .map(Optional::get)
+                    .collect(Collectors.toList());
+            String maybeMissingToolchains = missingToolchains.isEmpty()
+                    ? "some language versions"
+                    : String.format("the language versions=%s", missingToolchains);
+            log.error(
+                    "\n"
+                        + "\u001B[31m****************************************************************************************************\n"
```
crogoz: Logging with the red formatter, so it easy easy to spot:
Screenshot 2025-01-09 at 3 41 58 PM

gradle-jdks/src/main/java/com/palantir/gradle/jdks/flow/ToolchainFlowAction.java
```diff
@@ -0,0 +1,97 @@
...
+        ListProperty<String> getConfiguredJavaMajorVersions();
+    }
+
+    @Override
+    @SuppressWarnings("LineLength")
+    public void execute(Parameters parameters) {
+        parameters.getBuildResult().get().getFailure().ifPresent(failure -> {
+            List<Throwable> noToolchainsAvailable = Throwables.getCausalChain(failure).stream()
+                    .filter(throwable ->
+                            throwable instanceof org.gradle.jvm.toolchain.internal.NoToolchainAvailableException)
```
crogoz, line 58: this is an internal class, the testing-plugin would be useful here to make sure this workflow still works in more recent versions of gradle. 

I had to add the full qualified name of the  class here, because importing this class will fail the checkstyle.

gradle-jdks-setup/README.md
```diff
@@ -211,6 +211,28 @@ The plugin registers the following tasks:
...
+
+## Using a new java version
+
+We are only generating the jdk configuration files for the jdk versions that are used see: [jdkMajorVersionsToUse](https://github.com/palantir/gradle-jdks/blob/26d54605b7827b232024ec7e0c34a87983804492/gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdksExtension.java#L44)
+Using a new java version is a 2-step process: 
+1. Firstly, we need to generate the gradle jdk configuration files for the new java version that we are going to use
+2. Secondly, once the new files are generated, the javaVersion can be configured in the gradle build configuration.
+
+### Palantir specific, if using `baseline-java-version` through excavators
+1. Run `./gradlew generateGradleJdkConfigs --includeAllJdks`. The command will generate the Gradle jdk configuration files for all jdks.
```
CRogers: >all jdks the plugin is aware of.

gradle-jdks-setup/README.md
```diff
@@ -211,6 +211,28 @@ The plugin registers the following tasks:
...
+2. Update the javaVersion(s)/jdks extension values with the desired java major version.
+3. Run `./gradlew setupJdks` to only keep the Gradle jdk configuration files for the used Gradle JDKs.
+
+### Manual bumps of the java versions 
+1. Run `./gradlew generateGradleJdkConfigs --includeVersion=<newVersion> [--includeVersion=<otherNewVersion>]`. The command will generate the Gradle jdk configuration files for the major versions passed as parameters.
+2. Update the javaVersion(s)/jdks extension values with the desired java major version.
```
CRogers: We still need to run `./gradlew setupJdks` even with this approach as the user may stop using the old version, and we need to remove it

crogoz: Addressed, I thought I had the logic in the `check` task to test if there are any extra configuration files in gradle/jdks. I added now that logic in: https://github.com/palantir/gradle-jdks/pull/479/commits/1710e98961e625581cc0d668686f61f972f1d899#diff-53f3309ab573fe65d62f8b5bf40e9bb94478d82b4d6434cb6bcf9b25516813a4R134

crogoz:
`do not merge` until the excavators that bump the java versions have the new flag.

bmarcaur:
> `do not merge` until the excavators that bump the java versions have the new flag.

This has been merged in this PR: https://github.com/palantir/gradle-jdks/pull/525

crogoz:
Waiting on the [excavator PR](https://pl.ntr/2sG)

gradle-jdks-setup/README.md
```diff
@@ -211,6 +211,29 @@ The plugin registers the following tasks:
...
+
+### Palantir specific, if using `baseline-java-version` through excavators
+1. Run `./gradlew generateGradleJdkConfigs --includeAllJdks`. The command will generate the Gradle jdk configuration files for all jdks the plugin is aware of.
+2. Update the javaVersion(s)/jdks extension values with the desired java major version.
+3. Run `./gradlew setupJdks` to only keep the Gradle jdk configuration files for the used Gradle JDKs.
+
+### Manual bumps of the java versions 
+1. Run `./gradlew generateGradleJdkConfigs --includeVersion=<newVersion> [--includeVersion=<otherNewVersion>]`. The command will generate the Gradle jdk configuration files for the major versions passed as parameters.
+2. Update the javaVersion(s)/jdks extension values with the desired java major version.
+3. Run `./gradlew setupJdks` to only keep the Gradle jdk configuration files for the used Gradle JDKs.
```
felixdesouza:
```suggestion
We are only generating the JDK configuration files for the JDK versions that are used see: [`jdkMajorVersionsToUse`](https://github.com/palantir/gradle-jdks/blob/26d54605b7827b232024ec7e0c34a87983804492/gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdksExtension.java#L44)
Using a new Java version is a 2-step process: 
1. Firstly, we need to generate the `gradle-jdks` configuration files for the new Java version that we are going to use
2. Secondly, once the new files are generated, the `javaVersion` can be configured in the Gradle build configuration.

### Palantir specific, if using `baseline-java-version` through excavators
1. Run `./gradlew generateGradleJdkConfigs --includeAllJdks`. The command will generate the `gradle-jdks` configuration files for *all* JDKs the plugin is aware of.
2. Update the `javaVersion`(s)/`jdks` extension values with the desired Java major version.
3. Run `./gradlew setupJdks` to only keep the `gradle-jdks` configuration files for the used Gradle JDKs.

### Manual bumps of the java versions 
1. Run `./gradlew generateGradleJdkConfigs --includeVersion=<newVersion> [--includeVersion=<otherNewVersion>]`. The command will generate the `gradle-jdks` configuration files for the major versions passed as parameters.
2. Update the `javaVersion`(s)/`jdks` extension values with the desired Java major version.
3. Run `./gradlew setupJdks` to only keep the `gradle-jdks` configuration files for the used Gradle JDKs.
```

gradle-jdks-setup/README.md
```diff
@@ -211,6 +211,29 @@ The plugin registers the following tasks:
...
+
+## Using a new java version
+
+We are only generating the jdk configuration files for the jdk versions that are used see: [jdkMajorVersionsToUse](https://github.com/palantir/gradle-jdks/blob/26d54605b7827b232024ec7e0c34a87983804492/gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdksExtension.java#L44)
+Using a new java version is a 2-step process: 
+1. Firstly, we need to generate the gradle jdk configuration files for the new java version that we are going to use
+2. Secondly, once the new files are generated, the javaVersion can be configured in the gradle build configuration.
+
+### Palantir specific, if using `baseline-java-version` through excavators
+1. Run `./gradlew generateGradleJdkConfigs --includeAllJdks`. The command will generate the Gradle jdk configuration files for all jdks the plugin is aware of.
```
felixdesouza: Not sure if it will repeat/not make sense, but might be interesting to note how the plugin might be aware of the versions i.e. gradle-jdks-latest or whatever it's supposed to be

gradle-jdks/src/main/java/com/palantir/gradle/jdks/GradleJdksConfigs.java
```diff
@@ -44,6 +48,22 @@ public abstract class GradleJdksConfigs extends DefaultTask {
...
+            option = "includeAllJdks",
+            description = "Generates the configuration directories for all configured Java major versions.")
+    public abstract Property<Boolean> getIncludeAllJdks();
+
+    @Option(
+            option = "includeVersion",
+            description = "Generates the configuration directory for the java major versions passed in as parameters.")
+    public final void setIncludeVersionsFromCli(List<String> versions) {
+        getIncludeJavaMajorVersions().set(versions);
+    }
```
felixdesouza, line 62: do you necessarily need a separate method for this? couldn't you tack on `@Option` onto the `getIncludeJavaMajorVersions` property?

gradle-jdks/src/main/java/com/palantir/gradle/jdks/GradleJdksConfigs.java
```diff
@@ -73,8 +87,8 @@ protected abstract void applyGradleJdkFileAction(
     protected abstract void maybePrepareForAction(List<Path> targetPaths);
 
     public GradleJdksConfigs() {
-        // TODO(crogoz): change to false, once we switch to using jdksExtension's jdkMajorVersionsToUse
-        getIncludeAllJdks().convention(true);
+        getIncludeAllJdks().convention(false);
+        getIncludeJavaMajorVersions().convention(List.of());
```
felixdesouza: this should be empty list by default, so I don't think you need to set a convention?

gradle-jdks/src/main/java/com/palantir/gradle/jdks/GradleJdksConfigs.java
```diff
@@ -113,6 +127,19 @@ public final void action() {
                             + " `gradle-sls-docker` related plugins, see https://pl.ntr/2v1 to fix.");
         }
 
+        Set<String> configuredJavaMajorVersions = GradleJdksConfigsUtils.getConfiguredJavaMajorVersions(gradleJdksDir);
+        Set<String> unexpectedConfiguredJavaVersions = Sets.difference(
+                configuredJavaMajorVersions,
+                getJavaVersionToJdkDistros().get().keySet().stream()
+                        .map(JavaLanguageVersion::toString)
+                        .collect(Collectors.toSet()));
```
felixdesouza, line 134: won't this fail in the `--includeAllJdks` case? as `--includeVersion` property would be empty

gradle-jdks/src/main/java/com/palantir/gradle/jdks/GradleJdksConfigs.java
CRogers: Don't need to do this, the default is empty list.

gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdkDistributionConfigurator.java
```diff
@@ -36,8 +38,17 @@ public final class JdkDistributionConfigurator {
     private static final JavaLanguageVersion MINIMUM_SUPPORTED_JAVA_VERSION = JavaLanguageVersion.of(11);
 
     public static Map<JavaLanguageVersion, List<JdkDistributionConfig>> getJavaVersionToJdkDistros(
-            Project project, JdkDistributions jdkDistributions, JdksExtension jdksExtension) {
-        Set<JavaLanguageVersion> javaVersions = jdksExtension.getJdkMajorVersionsToUse().get().stream()
+            Project project,
+            JdkDistributions jdkDistributions,
+            JdksExtension jdksExtension,
+            Boolean includeAllJdks,
```
CRogers: should use a `boolean` instead of the boxed verison.

gradle-jdks/src/main/java/com/palantir/gradle/jdks/ToolchainsPlugin.java
CRogers: This is fine as internally we're almost all >=8.8

gradle-jdks/src/main/java/com/palantir/gradle/jdks/flow/ToolchainFlowAction.java
```diff
@@ -0,0 +1,97 @@
...
+                    .collect(Collectors.toList());
+            if (noToolchainsAvailable.isEmpty()) {
+                return;
+            }
+            List<String> missingToolchains = noToolchainsAvailable.stream()
+                    .map(exception -> {
+                        Pattern pattern = Pattern.compile("languageVersion=(\\d+)");
```
felixdesouza: nit: make a constant

gradle-jdks/src/main/java/com/palantir/gradle/jdks/flow/ToolchainFlowAction.java
```diff
@@ -0,0 +1,97 @@
...
+import org.gradle.api.flow.BuildWorkResult;
+import org.gradle.api.flow.FlowAction;
+import org.gradle.api.flow.FlowParameters;
+import org.gradle.api.provider.ListProperty;
+import org.gradle.api.provider.Property;
+import org.gradle.api.tasks.Input;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class ToolchainFlowAction implements FlowAction<Parameters> {
```
felixdesouza, line 37: this is very cool!

gradle-jdks/src/main/java/com/palantir/gradle/jdks/flow/ToolchainFlowAction.java
```diff
@@ -0,0 +1,97 @@
...
+                    : String.format("the java versions=%s", missingToolchains);
+            String includeVersionsOption = missingToolchains.isEmpty()
+                    ? "-includeVersion=<newJavaMajorVersion>"
+                    : missingToolchains.stream()
+                            .map(version -> String.format("--includeVersion=%s", version))
+                            .collect(Collectors.joining(" "));
+            log.error(
+                    "\n"
+                        + "\u001B[31m****************************************************************************************************\n"
+                        + "****************************************************************************************************\n"
```
felixdesouza: Nit: can we make it such that the text fits in the asterisks pls 🙈

gradle-jdks/src/test/groovy/com/palantir/gradle/jdks/GradleJdkPatcherIntegrationTest.groovy
```diff
@@ -20,6 +20,7 @@ package com.palantir.gradle.jdks
 import com.palantir.gradle.jdks.setup.common.CurrentArch
 import com.palantir.gradle.jdks.setup.common.CurrentOs
 import com.palantir.gradle.jdks.setup.common.GradleJdksPatchHelper
+import org.gradle.internal.impldep.com.amazonaws.util.Throwables
```
felixdesouza: don't think you want this

gradle-jdks/src/main/java/com/palantir/gradle/jdks/ToolchainsPlugin.java
```diff
@@ -58,12 +67,32 @@ public void apply(Project rootProject) {
...
-            rootProject
-                    .getExtensions()
-                    .getByType(BaselineJavaVersionsExtension.class)
-                    .getSetupJdkToolchains()
-                    .set(false);
+            BaselineJavaVersionsExtension baselineJavaVersionsExtension =
+                    rootProject.getExtensions().getByType(BaselineJavaVersionsExtension.class);
+            baselineJavaVersionsExtension.getSetupJdkToolchains().set(false);
+
+            jdksExtension.jdkMajorVersionsToUse().set(rootProject.provider(() -> Stream.of(
```
CRogers: Feels strange to `.set` here and `.add` below in the subprojects. I would probably just `.add` here too.

gradle-jdks/src/test/groovy/com/palantir/gradle/jdks/GradleJdkToolchainsIntegrationTest.groovy
```diff
@@ -254,6 +256,141 @@ class GradleJdkToolchainsIntegrationTest extends GradleJdkIntegrationSpec {
...
+        runTasksSuccessfully('wrapper')
+
+        when:
+        runTasksSuccessfully('generateGradleJdkConfigs', '--includeVersion=11')
+
+        then: 'generates directories for jdk version == 11'
+        Files.list(projectDir.toPath().resolve("gradle/jdks"))
+                .map { it -> it.getFileName().toString() }
+                .collect(Collectors.toList())
+                .containsAll(List.of("11"))
```
felixdesouza: would you not want it to be a stricter assertion i.e. it's *only* `11` or are there other jdks inserted? I can't fully tell yet

gradle-jdks/src/test/groovy/com/palantir/gradle/jdks/GradleJdkToolchainsIntegrationTest.groovy
```diff
@@ -19,6 +19,7 @@ package com.palantir.gradle.jdks
 import com.palantir.gradle.jdks.setup.common.CurrentArch
 import com.palantir.gradle.jdks.setup.common.CurrentOs
 import org.apache.commons.lang3.tuple.Pair
+import org.gradle.internal.impldep.com.amazonaws.util.Throwables
```
felixdesouza: can you use something from `Assertions` from assertJ instead?

gradle-jdks/src/main/java/com/palantir/gradle/jdks/ToolchainsPlugin.java
```diff
@@ -89,7 +118,11 @@ public void apply(Project rootProject) {
             task.getDaemonJavaVersion().set(jdksExtension.getDaemonTarget());
             task.getJavaVersionToJdkDistros()
                     .putAll(rootProject.provider(() -> JdkDistributionConfigurator.getJavaVersionToJdkDistros(
-                            rootProject, jdkDistributions, jdksExtension)));
+                            rootProject,
+                            jdkDistributions,
+                            jdksExtension,
+                            task.getIncludeAllJdks().get(),
+                            task.getIncludeJavaMajorVersions().get())));
```
felixdesouza, line 118: do you ever expect these to carry task dependencies?

if so, then probably want to do a zip instead

felixdesouza: oh, this is on the task itself, it's a bit confusing to grok this

gradle-jdks/src/main/java/com/palantir/gradle/jdks/ToolchainsPlugin.java
```diff
@@ -58,12 +67,32 @@ public void apply(Project rootProject) {
...
+                    .collect(Collectors.toSet())));
         });
+
+        rootProject.subprojects(
+                proj -> proj.getPluginManager().withPlugin("com.palantir.baseline-java-version", unused -> {
+                    BaselineJavaVersionExtension projectVersions =
+                            proj.getExtensions().getByType(BaselineJavaVersionExtension.class);
+
+                    jdksExtension.jdkMajorVersionsToUse().addAll(rootProject.provider(() -> Stream.of(
+                                    projectVersions.target(), projectVersions.runtime())
```
felixdesouza: hmm, I don't believe this needs to be a provider no?

felixdesouza: oh, it's nested

hmm, can we just have a small utility method that's like:

```java
private static Provider<JavaLanguageVersion> asJavaLanguageVersion(Property<ChosenJavaVersion> version) {
  return version.map(ChosenJavaVersion::javaLanguageVersion)
}
```

Then usage is just:

```java
jdksExtension.jdjkMajorVersionsToUse().addAll(
  asJavaLanguageVersion(projectVersions.target()),
  asJavaLanguageVersion(projectVersions.runtime())
)
```

Feels like it's a bit easier to see what's going on

gradle-jdks/src/main/java/com/palantir/gradle/jdks/ToolchainsPlugin.java
```diff
@@ -58,12 +67,32 @@ public void apply(Project rootProject) {
...
+            BaselineJavaVersionsExtension baselineJavaVersionsExtension =
+                    rootProject.getExtensions().getByType(BaselineJavaVersionsExtension.class);
+            baselineJavaVersionsExtension.getSetupJdkToolchains().set(false);
+
+            jdksExtension.jdkMajorVersionsToUse().set(rootProject.provider(() -> Stream.of(
+                            jdksExtension.getDaemonTarget(),
+                            baselineJavaVersionsExtension.libraryTarget(),
+                            baselineJavaVersionsExtension.runtime().map(ChosenJavaVersion::javaLanguageVersion),
+                            baselineJavaVersionsExtension
+                                    .distributionTarget()
```
CRogers: These are all properties, so we really should get into the habit of `.zip`ing properties together just in case there's a task that is used to calculate the value of the property (and by doing `rootProject.provider` we lose that dependency information).

Unfortunately this is really annoying to do as Gradle does not supply any sort of `Providers.zipWith(lots, of, different, values)` so best to hold off for now. I might make some util classes for this in gradle-utils.

gradle-jdks/src/main/java/com/palantir/gradle/jdks/ToolchainsPlugin.java
```diff
@@ -58,12 +67,32 @@ public void apply(Project rootProject) {
...
+                                    .map(ChosenJavaVersion::javaLanguageVersion))
```
felixdesouza: maybe something similar here to the subproject thing below

gradle-jdks/src/main/java/com/palantir/gradle/jdks/flow/ToolchainFlowAction.java
CRogers: This is (still) so slick!

gradle-jdks/src/main/java/com/palantir/gradle/jdks/flow/ToolchainFlowAction.java
```diff
@@ -0,0 +1,97 @@
...
+                    ? "some java versions"
+                    : String.format("the java versions=%s", missingToolchains);
+            String includeVersionsOption = missingToolchains.isEmpty()
+                    ? "-includeVersion=<newJavaMajorVersion>"
+                    : missingToolchains.stream()
+                            .map(version -> String.format("--includeVersion=%s", version))
+                            .collect(Collectors.joining(" "));
+            log.error(
+                    "\n"
+                        + "\u001B[31m****************************************************************************************************\n"
```
CRogers: Do we need `\u001B[31m` to change to red? I thought error logs were red in gradle at least?

fwiw I think if used on circle the `\u001B[31m` will look weird as it's not a terminal (hence perhaps best to let Gradle handle printing it or not if we can)

gradle-jdks/src/test/groovy/com/palantir/gradle/jdks/GradleJdkToolchainsIntegrationTest.groovy
CRogers: I think this should be the guava one

gradle-jdks/src/test/groovy/com/palantir/gradle/jdks/GradleJdkPatcherIntegrationTest.groovy
CRogers: Ditto for guava import

gradle-jdks/src/main/java/com/palantir/gradle/jdks/GradleJdksConfigs.java
crogoz: it fails if I do that with: `Command-line option '--includeVersion' does not take an argument.` when running the test in https://github.com/palantir/gradle-jdks/pull/479/files#diff-d8cb60820e935e74203f71da27d54ca1e390a3b5cb379b2a15f807309a3a4018R289

gradle-jdks/src/main/java/com/palantir/gradle/jdks/flow/ToolchainFlowAction.java
crogoz: I tried now without the `\u001B[31m` and it doesn't change it to red :(
Screenshot 2025-04-07 at 3 50 39 PM

CRogers: Ah, I'd at least only output the control characters when not on CI then.

gradle-jdks/src/main/java/com/palantir/gradle/jdks/ToolchainsPlugin.java
CRogers: Actually, why not just have lots of `jdksExtension.jdkMajorVersionsToUse().add/addAll(` calls here for each one, rather than trying to combine them together? This should do the trick just fine.

gradle-jdks/src/main/java/com/palantir/gradle/jdks/GradleJdksConfigs.java
crogoz: It will fail if you run `./gradlew generateGradleJdks --includeAllJdks` and then you run `./gradlew checkGradleJdks` - but this is expected. I want at the end of the build to only  keep the actual used JDKs. 

```
* What went wrong:
Execution failed for task ':checkGradleJdkConfigs'.
> Unexpected Java versions configured: [21]. Please run `./gradlew setupJdks` to regenerate the used JDKS.
```

gradle-jdks/src/main/java/com/palantir/gradle/jdks/flow/ToolchainFlowAction.java
crogoz: addressed this in https://github.com/palantir/gradle-jdks/pull/479/commits/4b8a2513668b586ad922dd6869f3e79c238f228b#diff-c7b32bc0b052f8a5cc11b4c0eed62a0c270c15a432e2a52dffe6bc0678df45f6R80: 
Screenshot 2025-04-08 at 12 09 33 PM

changelog/@unreleased/pr-479.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: fix
+fix:
+  description: [Gradle JDK Automanagement] Install only used JDKs
```
felixdesouza:
```suggestion
  description: '[Gradle JDK Automanagement] Install only used JDKs'
```

Changelog is failing otherwise

gradle-jdks/src/main/java/com/palantir/gradle/jdks/flow/ToolchainFlowAction.java
```diff
@@ -0,0 +1,108 @@
...
+                            maybeMissingToolchains,
+                            parameters.getConfiguredJavaMajorVersions().get()),
+                    "If you are trying to manually change the Java versions used, please follow the steps:",
+                    String.format(
+                            "\t- Make sure build.gradle files only use the configured java major versions: %s",
+                            parameters.getConfiguredJavaMajorVersions().get()),
+                    String.format(
+                            "\t- Run `./gradlew generateGradleJdkConfigs %s` to generate the jdk configuration files.",
+                            includeVersionsOption),
+                    "\t- Update the build.gradle's java versions with the newly configured jdks");
```
felixdesouza: I actually liked how you had a massive string before, it was a lot clearer, and I think you can still keep it.

I would just have a separate method that's like: `calculateHeaderFooter` which does what you already do.

Similarly I would have another method, `ansi` which takes in a string and does the filtering there. Then you can keep everything as one big string that gets formatted:

```
"""
%s%s
Gradle JDK Auto-management...

  - Make sure build.gradle files only use ...: %s
  ...
%s%s
""".formatted(ansi(ANSI_RED_COLOR), headerFooter, ..., headerFooter, ansi(ANSI_RESET_COLOR));
```

The ansi library is actually very nice for doing this sort of stuff, but imagine you don't want to bring in this dependency.

gradle-jdks/src/main/java/com/palantir/gradle/jdks/flow/ToolchainFlowAction.java
```diff
@@ -0,0 +1,114 @@
...
+            String headerFooter = "*".repeat(maxLineSize);
+
+            log.error(String.join(
+                    "\n",
+                    ImmutableList.<String>builder()
+                            .add(ansi(ANSI_RED_COLOR))
+                            .add(headerFooter)
+                            .add(explanation)
+                            .add(headerFooter)
+                            .add(ANSI_RESET_COLOR)
```
felixdesouza: missing `ansi` method here

CRogers (approved):

PR #493, opened 2025-01-07
Add default includeAllJdks option to generateGradleJdkConfigs

crogoz:
## After this PR

In preparation for https://github.com/palantir/gradle-jdks/pull/479/files#diff-53f3309ab573fe65d62f8b5bf40e9bb94478d82b4d6434cb6bcf9b25516813a4R79. That PR will only install & configure only the jdk versions used. Because of this, excavators that bump the java versions (eg. https://pl.ntr/2q1), would need to pre-install extra java versions. For this, I have added the command line option `./gradlew generateGradleJdkConfigs --includeAllJdks` that will generate all the jdk configuration files. 
In this PR, the addition is a noop. It only allows the excavators that update the java versions to use the new command line to ensure they are not broken when https://github.com/palantir/gradle-jdks/pull/479/files is merged. 

==COMMIT_MSG==
Add default includeAllJdks option to generateGradleJdkConfigs
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-493.v2.yml
+++ b/changelog/@unreleased/pr-493.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Add default includeAllJdks option to generateGradleJdkConfigs
+  links:
+  - https://github.com/palantir/gradle-jdks/pull/493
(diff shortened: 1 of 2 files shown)


CRogers:
:+1:

PR #534, opened 2025-05-06
Disable Gradle JDK Automanagement workflow when running on mingw

crogoz:
## Before this PR

Failing when using mingw shell: https://pl.ntr/2tn
```
ERROR Unsupported OS: MINGW64_NT-10.0-17763

ERROR: /c/Users/circleci/project/gradle/jdks/11//x86-64/local-path not found, aborting Gradle JDK setup

ERROR: /c/Users/circleci/project/gradle/jdks/11//x86-64/download-url not found, aborting Gradle JDK setup
```

## After this PR

Mingw is unsupported, Gradle JDK Automanagement will not be set up.
Tested successfully: https://pl.ntr/2to

==COMMIT_MSG==
[Gradle JDK Enablement] Mingw is not supported
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-534.v2.yml
+++ b/changelog/@unreleased/pr-534.v2.yml
@@ -1,5 +1,5 @@
 type: fix
 fix:
-  description: '[Gradle JDK Enablement] Mingw is not supported'
+  description: Disable Gradle JDK Automanagement workflow when running on mingw
   links:
   - https://github.com/palantir/gradle-jdks/pull/534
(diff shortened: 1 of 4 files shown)


changelog/@unreleased/pr-534.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: fix
+fix:
+  description: '[Gradle JDK Enablement] Mingw is not supported'
```
felixdesouza: Perhaps say something like 'Fallback to normal when using `mingw`' or replace `normal` with whatever it needs to be, idk

felixdesouza (approved):

PR #536, opened 2025-05-07
Allow `palantir-gradle-jdks` idea plugin dependency configuration via `idea-configuration plugin`

FinlayRJW:
## Before this PR
We had our own implementation to configure the external dependancy xml, this is now managed by the shared `idea-configuration` plugin

## After this PR
We use the `idea-configuration` plugin to add the dependancy on `palantir-gradle-jdks`

==COMMIT_MSG==
Allow `palantir-gradle-jdks` idea plugin dependency configuration via `idea-configuration plugin`
==COMMIT_MSG==


FinlayRJW:
hmm broken till bumped to java 17 see https://github.com/palantir/gradle-jdks/pull/428 - had a look and it seems that the issue is with this block 

```gradle
// DO NOT UPDATE! This needs to be kept as low as possible, such that it is compatible with all the jdk versions the
// consumers would want to set up with the JDK auto-management workflow (used by ./gradle-jdks-setup.sh and GradleWrapperMain.java classes)
javaVersion {
    target = 11
    runtime = 11
}
```

in `gradle-setup-jdks/build.gradle` I assume this is something we can't bump

crogoz (approved):

PR #547, opened 2025-05-14
Support Gradle 8.13

pkoenig10:
This internal type was renamed in https://github.com/gradle/gradle/pull/29166 and then again in https://github.com/gradle/gradle/pull/29376.

### DIFF
--- a/gradle-jdks/src/main/java/com/palantir/gradle/jdks/flow/ToolchainFlowAction.java
+++ b/gradle-jdks/src/main/java/com/palantir/gradle/jdks/flow/ToolchainFlowAction.java
@@ -22,6 +22,7 @@ import com.palantir.gradle.jdks.flow.ToolchainFlowAction.Parameters;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


gradle-jdks/src/main/java/com/palantir/gradle/jdks/flow/ToolchainFlowAction.java
```diff
@@ -37,6 +38,13 @@
 public final class ToolchainFlowAction implements FlowAction<Parameters> {
 
     private static final Logger log = LoggerFactory.getLogger(ToolchainFlowAction.class);
+
+    // This internal exception was renamed in https://github.com/gradle/gradle/pull/29166 and
+    // https://github.com/gradle/gradle/pull/29376
+    private static final Set<String> TOOLCHAIN_EXCEPTIONS = Set.of(
+            "org.gradle.jvm.toolchain.internal.NoToolchainAvailableException",
+            "org.gradle.jvm.toolchain.internal.install.exceptions.ToolchainProvisioningException");
```
pkoenig10, line 46: Is there an alternative solution that does not requiring using internal types?

CRogers: Briefly looking, I don't think so. This is definitely the less risky way of checking the exception, ~~although it does mean this feature may break without us knowing~~ this is what the tests in this repo should be checking.

CRogers: Hardcoded test versions strike again - will follow up with fixing that.

CRogers:
:+1:

PR #549, opened 2025-05-15
Refactor `is_arch_os_supported`

samialfhaily:
https://www.shellcheck.net/wiki/SC2091

### DIFF
--- a/gradle-jdks-setup/src/main/resources/gradle-jdks-setup.sh
+++ b/gradle-jdks-setup/src/main/resources/gradle-jdks-setup.sh
@@ -67,7 +67,7 @@ APP_GRADLE_DIR="$APP_HOME"/gradle
 # Loading gradle jdk functions
 . "$APP_GRADLE_DIR"/gradle-jdks-functions.sh
 
-if ! $(is_arch_os_supported); then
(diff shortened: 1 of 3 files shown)


gradle-jdks-setup/src/main/resources/gradle-jdks-setup.sh
```diff
@@ -67,7 +67,7 @@ APP_GRADLE_DIR="$APP_HOME"/gradle
 # Loading gradle jdk functions
 . "$APP_GRADLE_DIR"/gradle-jdks-functions.sh
 
-if ! $(is_arch_os_supported); then
+if ! eval "$(is_arch_os_supported)"; then
```
CRogers: I think we explicitly don't want the eval form; we just want to execute the function rather than run some code it produces

CRogers: Oh, it does actually `echo false` rather than `return 0/1`... We should probably change that up

samialfhaily: Made the change. Is this what you're thinking?

crogoz (approved):

PR #565, opened 2025-06-27
Make `gradle-jdks` configuration-cache compatible with the configuration phase

kelvinou01:
## Before this PR

We are rolling out the configuration cache [incrementally](https://github.com/palantir/gradle-incremental-configuration-cache). To get this plugin working on our repositories, we [need](https://github.com/palantir/gradle-incremental-configuration-cache?tab=readme-ov-file#limitations) to resolve all configuration phase issues, so `build` works, and the plugin can be merged. 

Running `./gradlew build --configuration-cache` with the plugin applied raises these errors:

```
- Class `com.palantir.gradle.jdks.setup.common.CommandRunner`: external process started 'security export -t certs -f pemseq -k /Library/Keychains/System.keychain'
  See https://docs.gradle.org/8.14.2/userguide/configuration_cache.html#config_cache:requirements:external_processes
```

The error arises from `PalantirCaPlugin::apply` calling [`security export -t certs -f pemseq -k /Library/Keychains/System.keychain`](https://github.com/palantir/gradle-jdks/blob/830d49486ea3e6021fb76dc213bfd3fd4d5d7b40/gradle-jdks/src/main/java/com/palantir/gradle/jdks/PalantirCaPlugin.java#L52). This call is done at configuration time, meaning the incremental rollout plugin is unable to be applied. 

## After this PR

By wrapping the external call in a value source (via providerFactory.exec), the error is resolved. 

==COMMIT_MSG==
Make `gradle-jdk` configuration-cache compatible for the configuration phase
==COMMIT_MSG==

## Possible downsides?

The external call and some logic surrounding it had to be duplicated, but I tried to keep duplication minimal anyways by refactoring intermediate processing into static utils methods.


gradle-jdks/src/test/groovy/com/palantir/gradle/jdks/ConfigurationCacheTest.groovy
```diff
@@ -0,0 +1,61 @@
...
+        given:
+        // language=Gradle
+        buildFile << '''
+            apply plugin: 'com.palantir.jdks.palantir-ca'
+
+            jdks {
+                daemonTarget = 21
+            }
+
+            def value = jdks.getCaCerts().get()
```
kelvinou01, line 39: As a note, the lazy + interconnected nature of Gradle plugins makes it difficult to test for regressions.

e.g. In the real world, `jdks.caCerts` is probably read by some other plugin. When `jdk.caCerts` is realised, it calls the external process, violating CC requirements in the config phase. 

Without manually tracing the plugin code and triggering CC violations, we have no idea if we're hitting all CC violating paths in a test.

gradle-jdks/src/main/java/com/palantir/gradle/jdks/GradleAwareCertificateSource.java
```diff
@@ -0,0 +1,87 @@
...
+public abstract class GradleAwareCertificateSource {
+
+    @Inject
+    protected abstract ProviderFactory getProviderFactory();
+
+    public final Optional<Provider<byte[]>> systemCertificates(ILogger logger) {
+        Os os = CurrentOs.get();
+        switch (os) {
+            case MACOS:
+                return Optional.of(macosSystemCertificates());
```
FinlayRJW: So I think that this will run the command every time this is called I believe the correct thing todo here is:

```java
public abstract class GradleAwareCertificateSource {

    private final Provider<...> macosSystemCert = macosSystemCertificates()

    public final Optional<Provider<byte[]>> systemCertificates(ILogger logger) {
        Os os = CurrentOs.get();
        switch (os) {
            case MACOS:
                return Optional.of(macosSystemCert);
        ...
    }
```
does that sound about right @felixdesouza?

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/CertificateSource.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public final class CertificateSource {
+    private CertificateSource() {}
+
+    public static Optional<byte[]> systemCertificates(ILogger logger) {
+        Os os = CurrentOs.get();
```
FinlayRJW, line 36: FLUP replace this with OperatingSystem / GradleOperationSystem from utils

FinlayRJW: hmm looks like CI is failing on the OS commands so might need todo this before we can merge here

gradle-jdks/src/main/java/com/palantir/gradle/jdks/GradleAwareCertificateSource.java
kelvinou01: Thanks for catching this. Let's keep only one provider for each routine, and return that provider for every call to `systemCertificate`.

FinlayRJW:
As discussed in 🧵 it would be better to simply not do the `PalantirCaPlugin` setup if that gradle property `palantir.jdk.setup.enabled` is set

kelvinou01:
> As discussed in 🧵 it would be better to simply not do the `PalantirCaPlugin` setup if that gradle property `palantir.jdk.setup.enabled` is set

Yes! Let's also add a docstring in `GradleJdksEnablement` to explain this behavior!

gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/CertificateSource.java
kelvinou01: Let's move this thread to GradleAwareCertificateSource, which is used by gradle!

gradle-jdks/src/main/java/com/palantir/gradle/jdks/GradleAwareCertificateSource.java
```diff
@@ -0,0 +1,98 @@
...
+    @Inject
+    protected abstract ProviderFactory getProviderFactory();
+
+    private Provider<OperatingSystem> osProvider;
+
+    private final Provider<byte[]> macosSystemCert = macosSystemCertificates();
+    private final Provider<byte[]> linuxSystemCert = linuxSystemCertificates();
+
+    @Inject
+    public GradleAwareCertificateSource(GradleOperatingSystem gradleOperatingSystem) {
```
kelvinou01, line 46: > FLUP replace this with OperatingSystem / GradleOperationSystem from utils

@FinlayRJW as per your suggestion, I've brought in `OperatingSystem` / `GradleOperationSystem` where needed for configuration cache to work.

As a FLUP, let's bump gradle-utils when the [fix](https://github.com/palantir/gradle-utils/pull/387) is out.

As a second FLUP, let's delete the `Os` class from this repo and replace all usages with `OperatingSystem` from utils.

kelvinou01:
@FinlayRJW @CRogers 

We've learned that `PalantirCaPlugin` (the root of our CC issues with gradle-jdks) can be avoided. It is legacy, and we can just [skip applying it](https://github.com/palantir/gradle-jdks-latest/pull/441) for repositories using [JDK Automanagement](https://github.com/palantir/gradle-jdks/issues/333). Repositories not using JDK Automanagement have bigger problems anyways, so let's not worry about them. 

I've verified that after manually bringing in changes from [baseline](https://github.com/palantir/gradle-baseline/pull/3140) and [gradle-jdks-latest](https://github.com/palantir/gradle-jdks-latest/pull/441), build with configuration cache works for some of our plugins. 

Let's close this PR.

PR #575, opened 2025-07-07
Switch to `gradle-utils`' `GradleOperatingSystem` to be configuration cache friendly

kelvinou01:
## Before this PR

We are rolling out the configuration cache [incrementally](https://github.com/palantir/gradle-incremental-configuration-cache). To get the rollout plugin working on our repositories, we [need](https://github.com/palantir/gradle-incremental-configuration-cache?tab=readme-ov-file#limitations) to resolve all configuration phase issues, so `build` works, and the plugin can be merged. 

`CurrentOs` does an external call to `ldd`, which breaks the configuration cache. By bringing in `GradleOperatingSystem` from gradle-utils, we unify our util usage and use a CC-friendly API for getting the OS. 

Blocked by https://github.com/palantir/gradle-utils/pull/397

## After this PR

==COMMIT_MSG==
Switch to `gradle-utils`' `GradleOperatingSystem` to be configuration cache friendly
==COMMIT_MSG==

## Possible downsides?

FLUP: Let's nuke `Arch` and switch to gradle-utils' `Architecture`


gradle-jdks-setup/build.gradle
```diff
@@ -5,6 +5,7 @@ apply plugin: 'org.unbroken-dome.test-sets'
 dependencies {
     // Avoid adding large compile dependencies here, as they will also be added in the gradle-jdks-setup-all*.jar using the fatJar task
     implementation project(':gradle-jdks-setup-common')
+    implementation 'com.palantir.gradle.utils:platform'
```
kelvinou01, line 8: This is small (only 4 classes) and brings in [very few dependencies](https://github.com/palantir/gradle-utils/blob/develop/platform/build.gradle)

gradle-jdks/src/main/java/com/palantir/gradle/jdks/SetupJdksTask.java
```diff
@@ -46,9 +48,29 @@ public abstract class SetupJdksTask extends DefaultTask {
     @Inject
     protected abstract ExecOperations getExecOperations();
 
+    @Inject
+    protected abstract ObjectFactory getObjectFactory();
+
+    /**
+     * If we make this an abstract getter and use @Nested, we get the following error:
```
kelvinou01: Note to self: In this PR, `@Nested` works for plugins but not tasks for some reason.  Maybe a bug w gradle?

gradle-jdks-settings/src/test/groovy/com/palantir/gradle/jdks/settings/ToolchainJdksSettingsPluginTest.groovy
```diff
@@ -97,7 +97,11 @@ class ToolchainJdksSettingsPluginTest extends IntegrationSpec {
 
         where:
         // testing for the different gradle versions to make sure the reflection in the settings plugin works
-        gradleVersionNumber << [GradleJdkTestUtils.GRADLE_7_6_VERSION, GradleJdkTestUtils.GRADLE_8_5_VERSION, GradleJdkTestUtils.GRADLE_8_8_VERSION]
+        gradleVersionNumber << [
+                GradleJdkTestUtils.GRADLE_7_6_4_VERSION,
```
kelvinou01, line 101: com.palantir.platform brings in Jackson 2.18.3 which needs Java 21, which needs Gradle 7.6.4, because it [brings support for multi-release jars with Java 21](https://docs.gradle.org/7.6.4/release-notes.html)

gradle-jdks-setup/build.gradle
CRogers: Why do we need this in gradle-jdks-setup? I thought we were continuing to use `CurrentOs`/`Os` there?

CRogers: Looks like it compiles without it.

gradle-jdks/src/main/java/com/palantir/gradle/jdks/GradleJdksConfigs.java
```diff
@@ -48,6 +52,13 @@ public abstract class GradleJdksConfigs extends DefaultTask {
     public static final String GRADLE_JDKS_SETUP_SCRIPT = "gradle-jdks-setup.sh";
     public static final String GRADLE_JDKS_FUNCTIONS_SCRIPT = "gradle-jdks-functions.sh";
 
+    @Inject
+    protected abstract ObjectFactory getObjectFactory();
+
+    private GradleOperatingSystem gradleOperatingSystem = getObjectFactory().newInstance(GradleOperatingSystem.class);
+
+    private final Provider<OperatingSystem> operatingSystem = gradleOperatingSystem.getOperatingSystem();
```
CRogers: Can we make a `@Internal @Nested GradleOperatingSystem` instead?

gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdksPlugin.java
```diff
@@ -17,25 +17,34 @@
...
 
-public final class JdksPlugin implements Plugin<Project> {
+public abstract class JdksPlugin implements Plugin<Project> {
 
     private static final Logger log = LoggerFactory.getLogger(JdksPlugin.class);
 
+    @Nested
+    protected abstract GradleOperatingSystem getOperatingSystem();
+
+    private Provider<OperatingSystem> osProvider = getOperatingSystem().getOperatingSystem();
```
CRogers: Can we just call `getOperatingSystem` directly where we need to?

gradle-jdks-settings/src/main/java/com/palantir/gradle/jdks/settings/ToolchainJdksSettingsPlugin.java
```diff
@@ -53,14 +55,21 @@
...
  */
-public final class ToolchainJdksSettingsPlugin implements Plugin<Settings> {
+public abstract class ToolchainJdksSettingsPlugin implements Plugin<Settings> {
 
     private static final Logger logger = Logging.getLogger(ToolchainJdksSettingsPlugin.class);
 
+    @Nested
+    protected abstract GradleOperatingSystem getGradleOperatingSystem();
+
+    private Provider<OperatingSystem> operatingSystem;
```
FinlayRJW: I don't think you should need todo this `gradle-utils` should store the result so it wont re-run the command each time its called - https://github.com/palantir/gradle-utils/blob/develop/platform/src/main/java/com/palantir/platform/GradleOperatingSystem.java#L31

gradle-jdks/src/main/java/com/palantir/gradle/jdks/SetupJdksTask.java
CRogers: I thought we managed to make this work - didn't it just need to be prefixed with `get`?

gradle-jdks-setup-common/src/main/java/com/palantir/gradle/jdks/setup/common/CurrentOs.java
```diff
@@ -23,6 +23,10 @@
 import java.util.concurrent.TimeUnit;
 import java.util.function.UnaryOperator;
 
+/**
+ * This class exists because gradle-jdks-setup has to be compatible with Java 11.
+ * If you don't need to support Java 11, prefer gradle-utils' <a href="https://github.com/palantir/gradle-utils/blob/17cacbf469772cef3c8ea213bd525f92ede6453f/platform/src/main/java/com/palantir/platform/OperatingSystem.java#L35">OperatingSystem</a>
```
FinlayRJW: ugh it's annoying we have to keep this around - would it be possible / worth making `gradle-utils:platform` be java 11 just so we can get rid of these? @CRogers thoughts?

gradle-jdks/src/main/java/com/palantir/gradle/jdks/SetupJdksTask.java
FinlayRJW: is this the thing where as it starts with `get` in gradle-utils it assumes its a managed property? - https://github.com/palantir/gradle-utils/blob/develop/platform/src/main/java/com/palantir/platform/GradleOperatingSystem.java#L33 .

Could we throw a PR up in `utils` to mark it internal?

CRogers: Looks like it really doesn't like `@Internal @Nested` when there is an input property inside the nested type:

```
- In plugin 'com.palantir.gradle.jdks.flow.ToolchainFailureFlowActionsPlugin_Decorated' type 'com.palantir.gradle.jdks.SetupJdksTask' property 'gradleOperatingSystem' annotated with @Internal should not be also annotated with @Nested.
    
    Reason: A property is ignored but also has input annotations.
    
    Possible solutions:
      1. Remove the input annotations.
      2. Remove the @Internal annotation.
``` 

Feels like bad behaviour from Gradle imo.

But even so, we can just add an `@Internal` on the type in gradle-utils. Then we can use it nicely here. If people really want the operating system to be a input, they can add their own `@Input` input property getter.

gradle-jdks-setup-common/src/main/java/com/palantir/gradle/jdks/setup/common/CurrentOs.java
CRogers: Even then, this setup jar is not run inside gradle so will never be compatible with `GradleOperatingSystem` as that requires Gradle to inject the object factory?

CRogers: Oh, or is the point it could call `OperatingSystem.get()` instead? If so, that would be fine, we can just bump down this platform jar to java to be compatible.

FinlayRJW: yeah I was thinking we would only use `OperatingSystem.get()` and then we could remove all of this

kelvinou01: If we were to make gradle-utils:platform java 11, we'd have to downgrade

- gradle-baseline from 6.40.0 to 6.5.0
- gradle-suppressible-error-prone from 2.11.0 to 2.5.0
- disable the gradle guide entirely

I don't think this an acceptable tradeoff? If gradle-jdks-setup needs more dependencies from our ecosystem, we'd have to downgrade more and more things. 

Let's keep a set of utils for java 11 within gradle-jdks-setup-common, and document that these utils are duplicated for java 11 use? Not sure if there's a better way.

gradle-jdks-settings/src/main/java/com/palantir/gradle/jdks/settings/ToolchainJdksSettingsPlugin.java
kelvinou01: I see, thanks for the tip.

gradle-jdks/src/main/java/com/palantir/gradle/jdks/JdksPlugin.java
kelvinou01: Yep!

gradle-jdks-setup/build.gradle
kelvinou01: Sorry forgot to remove this!

gradle-jdks/src/main/java/com/palantir/gradle/jdks/GradleJdksConfigs.java
kelvinou01: Yep, Ill make it nested, now that https://github.com/palantir/gradle-utils/pull/397 is out

gradle-jdks-setup-common/src/main/java/com/palantir/gradle/jdks/setup/common/CurrentOs.java
FinlayRJW: can use `sourceCompatibility` rather than `target` see https://github.com/palantir/gradle-utils/pull/398

kelvinou01: Nice! let's nuke CurrentOs and Os!

gradle-jdks/src/test/groovy/com/palantir/gradle/jdks/JdksExtensionProjectSpec.groovy
```diff
@@ -56,9 +68,9 @@ final class JdksExtensionProjectSpec extends IntegrationSpec {
                 jdks.jdkFor(JavaLanguageVersion.of(11), project).get().jdkFor(os).jdkFor(arch).jdkVersion.get()
             }
             
-            println('jdkVersion macos aarch64: ' + jdkVersionFor(Os.MACOS, Arch.AARCH64))
```
CRogers, line 59: Was a little worried that `BaselineJavaVersions` had this concept of `Os` leak into it, but it's fine, didn't see that jdkVersionFor is just defined for this test

CRogers (approved):

PR #577, opened 2025-07-10
Fix ABI break for `isGradleJdkSetupEnabled` by readding the deleted `isGradleJdkSetupEnabled(Path)` method

kelvinou01:
## Before this PR

In https://github.com/palantir/gradle-jdks/pull/575, we added an `OperatingSystem` parameter to `isGradleJdkSetupEnabled`, forgetting that it is a public library method being [used in other repositories](https://github.com/palantir/gradle-jdks-latest/blob/f94a936b8038d7a30684d112a775be9535057f28/gradle-jdks-latest/src/main/java/com/palantir/gradle/jdkslatest/LatestJdksPlugin.java#L35)

## After this PR

==COMMIT_MSG==
Fix ABI break for `isGradleJdkSetupEnabled` by readding the deleted `isGradleJdkSetupEnabled(Path)` method
==COMMIT_MSG==

### DIFF
--- a/gradle-jdks-enablement/src/main/java/com/palantir/gradle/jdks/enablement/GradleJdksEnablement.java
+++ b/gradle-jdks-enablement/src/main/java/com/palantir/gradle/jdks/enablement/GradleJdksEnablement.java
@@ -28,6 +28,10 @@ public final class GradleJdksEnablement {
 
     public static final String MINIMUM_SUPPORTED_GRADLE_VERSION = "7.6";
 
+    public static boolean isGradleJdkSetupEnabled(Path projectDir) {
+        return isGradleJdkSetupEnabled(OperatingSystem.get(), projectDir);
+    }
+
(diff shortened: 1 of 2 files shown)


gradle-jdks-enablement/src/main/java/com/palantir/gradle/jdks/enablement/GradleJdksEnablement.java
```diff
@@ -28,6 +28,10 @@ public final class GradleJdksEnablement {
 
     public static final String MINIMUM_SUPPORTED_GRADLE_VERSION = "7.6";
 
+    public static boolean isGradleJdkSetupEnabled(Path projectDir) {
```
FinlayRJW, line 31: Maybe we want to mark this with `@Deprecated` but not sure if that can block upgrades if users have `Xlint:deprecation` enabled - can be a FLUP if needed

FinlayRJW:
👍

PR #611, opened 2025-11-03
Remove `junit-bom` dependency

crogoz:
## After this PR

We shouldn't need to use boms, instead we should use the testImplementation junit-jupiter dep & testRuntimeOnly junit-platform-launcher dep. See [here](https://docs.gradle.org/current/userguide/java_testing.html#compiling_and_executing_junit_jupiter_tests)
==COMMIT_MSG==
Remove `junit-bom` dependency, add correct `junit-platform-launcher` dep for tests
==COMMIT_MSG==

### DIFF
--- a/gradle-jdks-distributions/build.gradle
+++ b/gradle-jdks-distributions/build.gradle
@@ -7,4 +7,6 @@ dependencies {
 
     testImplementation 'org.junit.jupiter:junit-jupiter'
     testImplementation 'org.assertj:assertj-core'
+
+    testRuntimeOnly 'org.junit.platform:junit-platform-launcher'
 }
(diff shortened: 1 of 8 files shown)


gradle-jdks-distributions/build.gradle
```diff
@@ -7,4 +7,6 @@ dependencies {
 
     testImplementation 'org.junit.jupiter:junit-jupiter'
     testImplementation 'org.assertj:assertj-core'
+
+    testRuntimeOnly 'org.junit.platform:junit-platform-launcher'
```
CRogers: I don't think these manually need to be added after https://github.com/palantir/gradle-baseline/pull/3053

PR #622, opened 2025-12-03
Use `javaVersions.allJavaVersionsUsed` where available instead of deeply reaching into baseline-java-versions

CRogers:
## Before this PR
See the related baseline-java-versions PR: https://github.com/palantir/gradle-baseline/pull/3399

`gradle-jdks` needs to know what JDKs are being used in a build to work out which JDKs it should actually set up. It currently does this by [deeply inspecting all the values inside the `javaVersions` and `javaVersion` extensions](https://github.com/palantir/gradle-jdks/blob/19d1f68fc8b933df9536d00b5d1641ade134a830/gradle-jdks/src/main/java/com/palantir/gradle/jdks/ToolchainsPlugin.java#L83-L97).

This is problematic, as [I'm about to add a `javaCompiler` option to `javaVersions`](https://github.com/palantir/gradle-baseline/pull/3342) it will need to know about. Currently, `gradle-jdks` is tightly coupled as it needs to know the internal structure of `baseline-java-versions`. When we come to add `javaCompiler`, if we set it to a JDK not already in used in a project (eg `javaCompiler = 25`) then `gradle-jdks` will not know to install it without it's own logic being updated.

If the structure of `baseline-java-versions` ever changes again, `gradle-jdks` will need updating, which is grim.

## After this PR
==COMMIT_MSG==
Use `javaVersions.allJavaVersionsUsed` where available instead of deeply reaching into baseline-java-versions
==COMMIT_MSG==

`gradle-jdks` now uses this method instead and avoid the tight coupling. When we add `javaCompiler`, it will get the new JDK automatically with no code changes.

## Possible downsides?
I've opted to implement this using reflection to see if the method is available, and if not falling back to the previous implementation. This is just so I can avoid bumping this plugin's baseline dependency to latest, which would block a lot of gradle-jdks upgrade PRs as there are repos that are stuck upgrading baseline at the moment. I'm also about to do a breaking baseline change, so want to reduce the scope of dependency upgrade problems for now.

Unfortunately, I can't test against the newer gradle-baseline without bumping the gradle-baseline dep. Or test against both the old and new codepath at the same time. So it's only testing against the old codepath. I have tested manually that the new codepath works though. When we bump the baseline dep, the new codepath will be tested/the old codepath will be removed.

### DIFF
--- a/gradle-jdks/src/main/java/com/palantir/gradle/jdks/ToolchainsPlugin.java
+++ b/gradle-jdks/src/main/java/com/palantir/gradle/jdks/ToolchainsPlugin.java
@@ -25,11 +25,15 @@ import com.palantir.gradle.jdks.GradleWrapperPatcher.GradleWrapperPatcherTask;
 import com.palantir.gradle.jdks.enablement.GradleJdksEnablement;
 import com.palantir.gradle.jdks.flow.ToolchainFailureFlowActionsPlugin;
 import com.palantir.platform.GradleOperatingSystem;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.Set;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.Task;
 import org.gradle.api.provider.Property;
 import org.gradle.api.provider.Provider;
+import org.gradle.api.provider.SetProperty;
 import org.gradle.api.tasks.Nested;
 import org.gradle.api.tasks.TaskProvider;
 import org.gradle.api.tasks.wrapper.Wrapper;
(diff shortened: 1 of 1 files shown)


gradle-jdks/src/main/java/com/palantir/gradle/jdks/ToolchainsPlugin.java
```diff
@@ -174,6 +183,24 @@ public final void apply(Project rootProject) {
         });
     }
 
+    private void gatherAllMajorVersionsUsedFromBaselineJavaVersions(
+            Project rootProject,
+            SetProperty<JavaLanguageVersion> jdkMajorVersionsToUse,
+            BaselineJavaVersionsExtension baselineJavaVersionsExtension) {
+
```
felixdesouza, line 190: you can effectively "freeze" this and not update it right? since when you release the new thing, it will have the `compiler` one and you it'll be automatically selected.

gradle-jdks/src/main/java/com/palantir/gradle/jdks/ToolchainsPlugin.java
```diff
@@ -78,23 +82,28 @@ public final void apply(Project rootProject) {
...
-            jdksExtension
-                    .jdkMajorVersionsToUse()
-                    .add(asJavaLanguageVersion(baselineJavaVersionsExtension.distributionTarget()));
+
+            try {
+                // We use reflection here to avoid having to bump baseline to latest to get this out.
+                // This is because there are quite a few stragglers and I'd rather not delay using this.
+                // Once enough time has passed and baseline is generally high enough, we should just bump
+                // the baseline dep and use the code directly.
+                Method allJavaVersionsUsed =
```
felixdesouza, line 94: not blocking as there are so few pieces ready rn, but I imagine with GMM, you could add a *constraint* for the version of baseline used, but not actually use the dependency if you did it based on variants. Although, I am not sure whether this would occur too late if the variant is calculated from a task and whether this is needed a lot sooner.

gradle-jdks/src/main/java/com/palantir/gradle/jdks/ToolchainsPlugin.java
CRogers: Yeah, this is just to handle the case that baseline hasn't upgraded yet, so it uses the old behaviour.

gradle-jdks/src/main/java/com/palantir/gradle/jdks/ToolchainsPlugin.java
CRogers: Yes, that would be ideal, and make it a `compileOnly` dep.

crogoz (approved):

PR #626, opened 2025-12-15
[Jdk Automanagement][generateGradleJdkConfigs] Add `onlyForCurrentOsArch` option

crogoz:
## After this PR

Adds an option `onlyForCurrentOsArch` when we are generating the jdk configuration files. The testing plugin only needs the jdk config files for the current os/arch. This simplifies the logic for finding the paths to the testing local jdks. 

==COMMIT_MSG==
[Jdk Automanagement] Preparing for gradle-plugin-testing integration
==COMMIT_MSG==

### DIFF
--- a/gradle-jdks/src/main/java/com/palantir/gradle/jdks/GradleJdksConfigs.java
+++ b/gradle-jdks/src/main/java/com/palantir/gradle/jdks/GradleJdksConfigs.java
@@ -122,8 +122,7 @@ public abstract class GradleJdksConfigs extends DefaultTask {
         getJavaVersionToJdkDistros().get().forEach((javaVersion, jdkDistros) -> {
             jdkDistros.forEach(jdkDistribution -> {
                 if (!getIncludeOnlyCurrentOsArch().get()
-                        || (getIncludeOnlyCurrentOsArch().get()
-                                && jdkDistribution.getOs().get().equals(os)
+                        || (jdkDistribution.getOs().get().equals(os)
                                 && jdkDistribution.getArch().get().equals(arch))) {
                     Path outputDir = gradleJdksDir
                             .resolve(javaVersion.toString())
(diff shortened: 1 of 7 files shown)


gradle-jdks/src/main/java/com/palantir/gradle/jdks/GradleJdksConfigs.java
```diff
@@ -65,6 +66,12 @@ public final void setIncludeVersionsFromCli(List<String> versions) {
         getIncludeJavaMajorVersions().set(versions);
     }
 
+    @Input
+    @Option(
+            option = "onlyForCurrentOsArch",
+            description = "Generates the configuration directories only for the current Os & Arch.")
+    public abstract Property<Boolean> getIncludeOnlyCurrentOsArch();
```
crogoz, line 73: example of usage in the gradle-plugin-testing invocation: https://github.com/palantir/gradle-plugin-testing/compare/cr/gradlew-gradleInvoker?expand=1#diff-4b13fa970a8965fd58bc7f42781597ec77437fba2dc9e021175c30ba85e5bf95R46

gradle-jdks/src/main/java/com/palantir/gradle/jdks/GradleJdksConfigs.java
```diff
@@ -112,14 +121,19 @@ public final void action() {
...
-                Path outputDir = gradleJdksDir
-                        .resolve(javaVersion.toString())
-                        .resolve(jdkDistribution.getOs().get().uiName())
-                        .resolve(jdkDistribution.getArch().get().uiName());
-                Path downloadUrlPath = outputDir.resolve("download-url");
-                Path localPath = outputDir.resolve("local-path");
-                applyGradleJdkFileAction(downloadUrlPath, localPath, jdkDistribution);
-                jdksDirectoryConfigured.set(true);
+                if (!getIncludeOnlyCurrentOsArch().get()
+                        || (getIncludeOnlyCurrentOsArch().get()
```
esword: nit - do you need this second call to (getIncludeOnlyCurrentOsArch().get() ?  If you are on this side of the || then you know that is already the case, right?

crogoz: yes, you are right! addressed in https://github.com/palantir/gradle-jdks/pull/626/commits/4ee9a51790bee1d4065d7ebdf8e621610c498370

esword (approved):

PR #661, opened 2026-01-21
[gradle-jdks-testing] Adding WithJdkAutomanagement annotation used by GradlePluginTests

crogoz:
## After this PR

Adding an annotation `WithJdkAutomanagement` that, when applied to `GradlePluginTests` test classes/tests, it will set-up all the necessary jdks & it will only allow running the gradle daemon & the tasks inside the tests using the configured jdks.  Follow-up for https://github.com/palantir/gradle-plugin-testing/pull/344.

More details [here](https://github.com/palantir/gradle-jdks/pull/661/changes#diff-2903068c3ad0954e6c77bcf6b7aa085169f58974220670de47db8f161e7829d3R1).

Modifications to the main JDK Auotmanagement workflow: 
- reverted the change that added "onlyForCurrentOsArch" (initial PR [here](https://github.com/palantir/gradle-jdks/pull/626)) because it is not needed anymore for this PR
- [in the test mode ](https://github.com/palantir/gradle-jdks/pull/661/changes#diff-d14af7703cef8b606999ab3937204e21f209c24a65a6196c2c63b24ac0fd8c40R188), we will skip running `./gradlew javaToolchains` as part of the `setupJdks` task. This is because running a `./gradlew` from the Testing framework will not propagate the classpaths, so the plugins used by the tests won't be found (internal [thread](https://pl.ntr/2zI))

==COMMIT_MSG==
Adding `WithJdkAutomanagement` annotation used by GradlePluginTests
==COMMIT_MSG==


gradle-jdks-testing/src/main/java/com/palantir/gradle/jdks/testing/GradleWithJdksInvoker.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    @Override
+    public GradleVersion gradleVersion() {
+        return delegate.gradleVersion();
+    }
+
+    @Override
+    public GradleInvocation withArgs(String... args) {
+        GradleInvocation setupJdkManagement = delegate.withArgs("wrapper", "setupJdks");
```
CRogers: Do we need to call `wrapper` here? I think this should only matter if we need to patch the `./gradlew` script which only matters if we actually use `./gradlew`?

gradle-jdks-testing/src/test/java/com/palantir/gradle/jdks/testing/WithJdkAutomanagementTests.java
```diff
@@ -0,0 +1,74 @@
...
+import com.palantir.gradle.testing.junit.DecoratorContext;
+import com.palantir.gradle.testing.junit.RegistersGradleInvokerDecorator;
+import com.palantir.gradle.testing.project.RootProject;
+import java.nio.file.Path;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Tests for the {@link WithJdkAutomanagement} annotation and its decorator factory.
+ */
+class WithJdkAutomanagementTests {
```
CRogers: I don't think these tests are that useful imo, it's just mocking stuff out. Better to have real integration tests that test against junit to check it's working properly.

gradle-jdks-testing/src/main/java/com/palantir/gradle/jdks/testing/GradleWithJdksInvoker.java
```diff
@@ -0,0 +1,116 @@
...
+    private static String readLocalPath(Path path) {
+        try {
+            return Files.readString(path).trim();
+        } catch (IOException e) {
+            throw new UncheckedIOException(String.format("Failed to read the path %s", path), e);
+        }
+    }
+
+    private static void setupRootProject(RootProject rootProject) {
+        rootProject.gradlePropertiesFile().appendProperty("palantir.jdk.setup.enabled", "true");
```
CRogers: Does this append the property multiple times if Gradle is invoked multiple times?

gradle-jdks-testing/src/main/java/com/palantir/gradle/jdks/testing/JdkSetupFailureException.java
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.gradle.jdks.testing;
+
+/**
+ * Exception thrown when JDK automanagement setup fails.
+ */
+public final class JdkSetupFailureException extends RuntimeException {
```
CRogers: Does this need to be public? Are we expecting people to explicitly catch it?

crogoz: I caught the exception [here](https://github.com/palantir/gradle-jdks/pull/661/changes#diff-76cd07fe207bbe1a20c1963530d1ada1e0e1a26970280cc81bb38fbf427a6179R103) - which I suspect would be the case for the follow-up PR where I would migrate the tests to this new workflow.
But I guess I can also only check the error message.

gradle-jdks-testing/src/main/java/com/palantir/gradle/jdks/testing/GradleWithJdksInvoker.java
crogoz: I don't think so, I am only setting up the rootProject in the constructor of the GradleInvoker, but I'll add a test for this

gradle-jdks-testing/src/main/java/com/palantir/gradle/jdks/testing/GradleWithJdksInvoker.java
crogoz: I needed it since "setupJdks" requires the gradlew patched script to be present. I can modify setupJkds such that this file is optional or I could split setupJdks in 2 tasks - 1 that does the setup of the jdks and the second one that runs "./gradlew javaToolchains".

crogoz: addressed it in https://github.com/palantir/gradle-jdks/pull/661/changes/48943badb0d3d5fb843e208e92af4d45df9951fa. Went with the @Optional gradlew file, let me know what you think

gradle-jdks-testing/src/main/java/com/palantir/gradle/jdks/testing/GradleWithJdksInvoker.java
```diff
@@ -0,0 +1,110 @@
...
+
+    GradleWithJdksInvoker(RootProject rootProject, GradleInvoker delegate) {
+        this.rootProject = rootProject;
+        this.delegate = delegate;
+        setupRootProject(rootProject);
+    }
+
+    @Override
+    public GradleInvocation withArgs(String... args) {
+        GradleInvocation setupJdkManagement = delegate.withArgs("setupJdks");
```
CRogers: I wonder if we actually want to call `setupJdks` before each command or whether we just want to do it in something like `beforeEach`? We wouldn't call `setupJdks` before each call in regular Gradle runs - I'm wondering whether this might actually hide something that might be interesting to see in tests - something like changing the jdks, then having to run `setupJdks` to get it to work

CRogers: I guess an issue here is that there may be other `@BeforeEach` which could invoke gradle that our `BeforeEachExtension` might not run before - although we could probably set this up so it does?

CRogers: I think we might just need to `gradle/gradle-jdks-setup.sh` like `gradlew` and IntelliJ do?

crogoz: > I think we might just need to gradle/gradle-jdks-setup.sh like gradlew and IntelliJ do?

`setupJdks` will only run `gradle/gradle-jdks-setup.sh` in any case when in testing mode: see [here](https://github.com/palantir/gradle-jdks/pull/661/changes#diff-874bce15e4b615599babbfed22b0a6b328b3e65dcfaef2acf53a28a29f07e7f5R66). This way the "setupJdks" is in charge of the processBuilder & reading the output etc.

gradle-jdks-testing/src/main/java/com/palantir/gradle/jdks/testing/GradleWithJdksInvoker.java
crogoz: opened https://github.com/palantir/gradle-plugin-testing/pull/378. I need to setup the rootProject [here](https://github.com/palantir/gradle-jdks/pull/661/changes#diff-d94e3df4c46ffd0bd4bfb7c2b432b13c519268d7978f385acdc82ed7567b03e3R60) before each invoker invocation for cases when in tests we have `buildGradle().overwrite(..)` to make sure the Jdks setup is still maintained.

crogoz:
depends on https://github.com/palantir/gradle-plugin-testing/pull/378

gradle-jdks-testing/src/main/java/com/palantir/gradle/jdks/testing/GradleWithJdksInvoker.java
```diff
@@ -0,0 +1,117 @@
...
+    GradleWithJdksInvoker(ExtensionContext extensionContext, RootProject rootProject, GradleInvoker delegate) {
+        this.extensionContext = extensionContext;
+        this.rootProject = rootProject;
+        this.delegate = delegate;
+    }
+
+    @Override
+    public GradleInvocation with(Options options) {
+        setupOrCheckRootProject(rootProject);
+        if (!GradleWithJdksStore.hasRunJdksSetup(extensionContext)) {
```
crogoz, line 61: we are only running setupJdks once for every context, this allows tests to modify the `jdks` if they need to.

gradle-jdks/src/main/java/com/palantir/gradle/jdks/GradleJdksConfigs.java
```diff
@@ -66,12 +66,6 @@ public final void setIncludeVersionsFromCli(List<String> versions) {
         getIncludeJavaMajorVersions().set(versions);
     }
 
-    @Input
-    @Option(
-            option = "onlyForCurrentOsArch",
-            description = "Generates the configuration directories only for the current Os & Arch.")
-    public abstract Property<Boolean> getIncludeOnlyCurrentOsArch();
```
crogoz, line 73: reverted the PR that introduced `onlyForCurrentOsArch`. Not required anymore by the testing plugin since we are finding the correct jdk paths here: https://github.com/palantir/gradle-jdks/pull/661/changes#diff-d94e3df4c46ffd0bd4bfb7c2b432b13c519268d7978f385acdc82ed7567b03e3R79

gradle-jdks/src/main/java/com/palantir/gradle/jdks/ToolchainsPlugin.java
```diff
@@ -172,11 +173,24 @@ public final void apply(Project rootProject) {
...
-            setupJdksTask.getGradlewScript().set(wrapperPatcherTask.get().getPatchedGradlewScript());
-            setupJdksTask.dependsOn(generateGradleJdkConfigs, wrapperPatcherTask);
+            setupJdksTask
+                    .getGradleJdksSetupScript()
+                    .fileProvider(generateGradleJdkConfigs.map(task -> task.getOutputGradleDirectory()
+                            .file("gradle-jdks-setup.sh")
+                            .get()
+                            .getAsFile()));
+
+            if (!environmentVariables.isInTestMode().get()) {
```
crogoz, line 188: when running in testMode we shouldn't run the `./gradlew` script as it will lead to issues finding the plugins in the test.

gradle-jdks-testing/src/main/java/com/palantir/gradle/jdks/testing/GradleWithJdksInvocation.java
```diff
@@ -0,0 +1,64 @@
...
+    @Override
+    public InvocationResult buildsSuccessfully() {
+        setupJdkAutomanagement();
+        try {
+            return tasksInvocation.call().buildsSuccessfully();
+        } catch (UnexpectedBuildFailure e) {
+            throw e;
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
```
CRogers: Why do we need to try-catch and rethrow these? `UnexpectedBuildFailure` is unchecked, `Exception` is checked by not in the `throws` so should never appear?

gradle-jdks-testing/src/main/java/com/palantir/gradle/jdks/testing/GradleWithJdksInvoker.java
```diff
@@ -0,0 +1,117 @@
...
+
+    GradleWithJdksInvoker(ExtensionContext extensionContext, RootProject rootProject, GradleInvoker delegate) {
+        this.extensionContext = extensionContext;
+        this.rootProject = rootProject;
+        this.delegate = delegate;
+    }
+
+    @Override
+    public GradleInvocation with(Options options) {
+        setupOrCheckRootProject(rootProject);
```
CRogers, line 60: I wonder whether this is worth doing precisely once too, in case we wanted to actually change on of these parameters in tests?

gradle-jdks-testing/src/main/java/com/palantir/gradle/jdks/testing/JdkAutomanagementDecorator.java
```diff
@@ -0,0 +1,34 @@
...
+import com.palantir.gradle.testing.execution.GradleInvoker;
+import com.palantir.gradle.testing.junit.DecoratorContext;
+import com.palantir.gradle.testing.junit.GradleInvokerDecorator;
+import java.lang.annotation.Annotation;
+import java.util.List;
+
+/**
+ * Decorator that enables JDK automanagement for Gradle invocations.
+ */
+public final class JdkAutomanagementDecorator implements GradleInvokerDecorator<Annotation> {
```
CRogers: Might as well make this `GradleInvokerDecorator<WithJdkAutomanagement>` as we only have one annotation, right?

gradle-jdks-testing/src/main/java/com/palantir/gradle/jdks/testing/GradleWithJdksInvoker.java
crogoz: I don't think it would work atm, as any setup in the code that does `buildGradle().overwrite()` will remove the JDK automanagement setup and it will lead to weird hard-to-debug failures.

CRogers:
:+1:

CRogers:
:+1:

crogoz:
👍

PR #693, opened 2026-02-20
[GradleWithJdksInvocation] SetupJdkAutoManagement builds expectations

crogoz:
## Before this PR

When trying to move the tests in gradle-baseline to the new Decorator, when setupJdks would fail and .buildWithFailure() would be called, we would throw an exception: 

https://github.com/palantir/gradle-baseline/pull/3504/changes#diff-06ccfb29d0444750df9fb7383f40f9ffa8d4766be2dc82211fa122e569fc3a8bL545-R553.

## After this PR
Switched `RuntimeException` to  `JdkSetupFailureException` in the tasks that are setting up the jdk automanagement. 

When `setupJdks` fails, check if the exception type is either`JdkSetupFailureException` or  `ToolchainProvisioningException`. In those cases, throw a `JdkSetupFailureException` - because the error is related to the jdk setup.

==COMMIT_MSG==
[GradleWithJdksInvocation] SetupJdkAutoManagement builds expectations
==COMMIT_MSG==


gradle-jdks-testing/src/main/java/com/palantir/gradle/jdks/testing/GradleWithJdksInvocation.java
```diff
@@ -29,7 +30,12 @@ record GradleWithJdksInvocation(
 
     @Override
     public InvocationResult buildsSuccessfully() {
-        setupJdkAutomanagement();
+        try {
+            setupJdkAutoManagement();
+        } catch (UnexpectedBuildFailure e) {
+            throw new JdkSetupFailureException(e);
```
crogoz: I think wrapping this exception in `JdkSetupFailureException` will make it more obvious that the failure is at the `setupJdk` logic.

However, this could also happen when `build.gradle` is wrongly set up (getting a `GradleScriptException`). I am not sure though how to differentiate between the user build.gradle errors and the build.gradle setupJdks errors

gradle-jdks-testing/src/main/java/com/palantir/gradle/jdks/testing/GradleWithJdksInvocation.java
```diff
@@ -39,20 +45,21 @@ public InvocationResult buildsSuccessfully() {
 
     @Override
     public InvocationResult buildsWithFailure() {
-        setupJdkAutomanagement();
+        try {
+            setupJdkAutoManagement();
+        } catch (UnexpectedBuildFailure e) {
+            return setupInvocation.buildsWithFailure();
```
crogoz: `buildsWithFailure()` might mean that either the `setupJdks` is failing or the actual `tasksInvocation` fails. 
If it fails at `setupJdks` stage, we will just return the result of that invocation.

gradle-jdks/src/test/java/com/palantir/gradle/jdks/GradleWithJdksTest.java
```diff
@@ -108,8 +108,15 @@ void fails_when_jdk_is_not_configured(GradleInvoker invoker, RootProject rootPro
...
             """);
+
         assertThatThrownBy(() -> invoker.withArgs("javaToolchains").buildsSuccessfully())
                 .hasMessageContaining("Gradle daemon JDK version is `24` but no JDK configured for that version.");
+
+        InvocationResult result = invoker.withArgs("javaToolchains").buildsWithFailure();
+        result.assertThat()
+                .output()
+                .contains("Gradle daemon JDK version is `24` but no JDK configured for that version.");
+        result.assertThat().task("javaToolchains").notOnTaskGraph();
```
CRogers: I think we need a couple more tests, at least:

1. Where there is a an error in a build script (eg throw exception), both with `buildsSuccessfully` and `buildsWithFailure`.
2. Error when running a some task with `buildSuccessfully`

These tests might make it easier to work out if we can differentiate between the JDK infra issues and issues with the rest of the build the user has made.

gradle-jdks-testing/src/main/java/com/palantir/gradle/jdks/testing/GradleWithJdksInvocation.java
CRogers: > I am not sure though how to differentiate between the user build.gradle errors and the build.gradle setupJdks errors

I think we can at least search for common strings or exception types in the error messages for jdks, [like we do for configuration cache](https://github.com/palantir/gradle-plugin-testing/blob/8a2e862ddd0bf71b3dda369208579e283be910a3/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/ConfigurationCacheInvocation.java#L49)?

gradle-jdks-testing/src/main/java/com/palantir/gradle/jdks/testing/GradleWithJdksInvocation.java
```diff
@@ -16,43 +16,74 @@
...
         try {
-            return tasksInvocation.call().buildsSuccessfully();
-        } catch (Exception e) {
-            throw new RuntimeException(e);
+            setupInvocation.buildsSuccessfully();
+            try {
+                // we expect this call to be successful if `setupInvocation` was successful.
+                javaToolchainsInvocation.call().buildsSuccessfully();
+            } catch (Exception e) {
+                throw new JdkSetupFailureException(e);
```
crogoz: I think it is safe to assume that id `setupInvocation` was successful and `javaToolchains` has failed, the reason was a jdk setup misconfiguration.

gradle-jdks-testing/src/main/java/com/palantir/gradle/jdks/testing/GradleWithJdksInvocation.java
```diff
@@ -16,43 +16,74 @@
...
     public InvocationResult buildsSuccessfully() {
-        setupJdkAutomanagement();
         try {
-            return tasksInvocation.call().buildsSuccessfully();
-        } catch (Exception e) {
-            throw new RuntimeException(e);
+            setupInvocation.buildsSuccessfully();
+            try {
+                // we expect this call to be successful if `setupInvocation` was successful.
+                javaToolchainsInvocation.call().buildsSuccessfully();
```
crogoz: calling `javaToolchainsInvocation` to simulate the normal `setupJdks` that would normally call `./gradlew javaToolchains`. 

This is a separate invocation because we need to set the `gradle.java.home` which is retrieved after the `setupJdks` [call](https://github.com/palantir/gradle-jdks/pull/693/changes#diff-d94e3df4c46ffd0bd4bfb7c2b432b13c519268d7978f385acdc82ed7567b03e3R89)

gradle-jdks-testing/src/main/java/com/palantir/gradle/jdks/testing/GradleWithJdksInvocation.java
```diff
@@ -16,43 +16,74 @@
...
+            } catch (Exception e) {
+                throw new JdkSetupFailureException(e);
+            }
+            markSetupComplete.run();
+        } catch (UnexpectedBuildFailure e) {
```
CRogers, line 48: I would make this try-catch be minimally around the setupInvocation. Currently it looks like it might apply to the `javaToolchains` one.

gradle-jdks-testing/src/main/java/com/palantir/gradle/jdks/testing/GradleWithJdksInvocation.java
```diff
@@ -16,43 +16,74 @@
...
-            return tasksInvocation.call().buildsWithFailure();
+            setupInvocation.buildsSuccessfully();
+            try {
+                // we expect this call to be successful if `setupInvocation` was successful.
+                javaToolchainsInvocation.call().buildsSuccessfully();
+            } catch (Exception e) {
+                return runWithFailure(javaToolchainsInvocation);
+            }
+            markSetupComplete.run();
+        } catch (UnexpectedBuildFailure e) {
```
CRogers, line 60: This try-catch should also be minimally around the `setupInvocation`.

I'm also a bit confused about this code. We run `setupInvocation` expecting it to succeed. If it fails, we run it again expecting it to fail? Why do we run it again?

I also would have thought we would want to separate out the "failing because JDK setup issues" vs "failing because of user code entered into the test that we expected to fail"

gradle-jdks-testing/src/main/java/com/palantir/gradle/jdks/testing/GradleWithJdksInvocation.java
```diff
@@ -16,43 +16,74 @@
...
+
+        return runWithFailure(tasksInvocation);
+    }
+
+    private InvocationResult runWithFailure(Callable<GradleInvocation> invocation) {
+        try {
+            return invocation.call().buildsWithFailure();
         } catch (Exception e) {
             throw new RuntimeException(e);
         }
```
CRogers: Is there a point in this? I don't think there is checked exception to catch here, so I think this is just wrapping any and all exceptions in `RuntimeException`s, which I don't think we need to do? We can just use the original runtime exception that is thrown?

Same in the method below.

crogoz: there is a checked exception - call() can throw an `Exception`. I've modified in https://github.com/palantir/gradle-jdks/pull/693/commits/057f0f17489bd7688c78873f7c78c43b13a5a0df to simply throw `RuntimeExceptions`

gradle-jdks-testing/src/main/java/com/palantir/gradle/jdks/testing/GradleWithJdksInvocation.java
crogoz: > We run setupInvocation expecting it to succeed. If it fails, we run it again expecting it to fail? Why do we run it again?

I'd need to make a public constructor for the InvocationResult. Opened a PR for this [here](https://github.com/palantir/gradle-plugin-testing/pull/393)

> I also would have thought we would want to separate out the "failing because JDK setup issues" vs "failing because of user code entered into the test that we expected to fail"

This is something that I am doing in `buildSuccessfully()`, it throws JdkSetupFailure if it failed because of JDK setup issues. 

For the `buildWithFailure` - do we want to throw in case of JdkSetupFailure ? this is the current case atm  - [see gradle-baseline PR](https://github.com/palantir/gradle-baseline/pull/3504/changes#diff-06ccfb29d0444750df9fb7383f40f9ffa8d4766be2dc82211fa122e569fc3a8bR552). With the changes I've made in this PR, the build will simply fail.

gradle-jdks-testing/src/main/java/com/palantir/gradle/jdks/testing/GradleWithJdksInvocation.java
```diff
@@ -16,43 +16,76 @@
...
+    private static GradleInvocation call(Callable<GradleInvocation> invocation) {
         try {
-            setupInvocation.buildsSuccessfully();
-            markSetupComplete.run();
+            return invocation.call();
+        } catch (RuntimeException e) {
+            throw e;
         } catch (Exception e) {
-            throw new JdkSetupFailureException(e);
+            throw new RuntimeException(e);
```
CRogers: To avoid the checked exception, can we use `Consumer` instead of `Callable`?

gradle-jdks-testing/src/main/java/com/palantir/gradle/jdks/testing/GradleWithJdksInvocation.java
CRogers: Ah, I see the issue with making the `InvocationResult`.

> For the buildWithFailure - do we want to throw in case of JdkSetupFailure ?

I think we do - if JDK setup has failed, I think it should be an unexpected failure - not the failure the test author was expecting they created. If the test author was expecting a JDK setup failure, they can catch it manually?

gradle-jdks-testing/src/main/java/com/palantir/gradle/jdks/testing/GradleWithJdksInvocation.java
```diff
@@ -16,43 +16,76 @@
...
-            throw new RuntimeException(e);
+            return runWithFailure(javaToolchainsInvocation);
         }
+        markSetupComplete.run();
+
+        return runWithFailure(tasksInvocation);
     }
 
-    private void setupJdkAutomanagement() {
+    private InvocationResult runWithFailure(Callable<GradleInvocation> invocation) {
```
CRogers: Not sure we'll need this extra indirection if we get rid of checked exceptions.

gradle-jdks-testing/src/main/java/com/palantir/gradle/jdks/testing/GradleWithJdksInvocation.java
crogoz: I've changed the logic a bit such that I am checking for both `JdkSetupFailureException` or `ToolchainProvisioningException` presence. If any of those are present, then I will throw the `JdkSetupFailureException`

gradle-jdks-testing/src/main/java/com/palantir/gradle/jdks/testing/GradleWithJdksInvocation.java
```diff
@@ -16,43 +16,60 @@
...
-            throw new JdkSetupFailureException(e);
+        } catch (UnexpectedBuildFailure e) {
+            if (e.getMessage().contains("com.palantir.gradle.jdks.setup.JdkSetupFailureException:")
+                    || e.getMessage().contains("ToolchainProvisioningException:")) {
+                throw new JdkSetupFailureException(e.getBuildResult().getOutput());
+            }
+            return InvocationResult.from(e);
+        }
+        InvocationResult result = tasksInvocation.get().buildsWithFailure();
+        if (result.output().contains("ToolchainProvisioningException:")) {
```
crogoz, line 68: test case here: https://github.com/palantir/gradle-jdks/pull/693/changes#diff-76cd07fe207bbe1a20c1963530d1ada1e0e1a26970280cc81bb38fbf427a6179R161

CRogers (approved):

PR #708, opened 2026-03-02
Added custom assertion for JDK directories

felixdesouza:
## Before this PR
Assertions declared in methods lose composability and also lose nice fluent descriptions. It also trips up the AI because they do the weird comment on top because as far as it can tell, it's just an arbitrary string you're passing through.

```java
// the extra directory was deleted
assertJdkDirectories(rootProject, Set.of("11", DAEMON_MAJOR_VERSION_17), "the extra directory was deleted");
```

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

I got claude to generate a `JdkDirectoriesAssert` that encapsulates some of that. It was relatively straightforward to generate and relatively low cost in time and effort, that it feels like a no-brainer to do. Here's how you use it:

```java
assertThatJdkDirectories(rootProject)
                .as("generates directories for jdk version == 11, 17")
                .containsExactJdks(11, 17);
```

### DIFF
--- a/gradle-jdks/src/test/java/com/palantir/gradle/jdks/GradleJdkPatcherIntegrationTest.java
+++ b/gradle-jdks/src/test/java/com/palantir/gradle/jdks/GradleJdkPatcherIntegrationTest.java
@@ -16,6 +16,7 @@
 
 package com.palantir.gradle.jdks;
 
+import static com.palantir.gradle.jdks.JdkDirectoriesAssert.assertThatJdkDirectories;
 import static com.palantir.gradle.testing.assertion.GradlePluginTestAssertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
@@ -29,15 +30,7 @@ import com.palantir.gradle.testing.junit.DisabledConfigurationCache;
 import com.palantir.gradle.testing.junit.GradlePluginTests;
 import com.palantir.gradle.testing.project.RootProject;
 import com.palantir.platform.OperatingSystem;
-import java.io.IOException;
-import java.io.UncheckedIOException;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.util.Set;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 import org.apache.commons.lang3.tuple.Pair;
-import org.assertj.core.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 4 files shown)


gradle-jdks/src/test/java/com/palantir/gradle/jdks/GradleJdkPatcherIntegrationTest.java
```diff
@@ -199,20 +192,14 @@ void no_gradle_wrapper_patch_if_jdk_setup_not_enabled(GradleInvoker gradle, Root
...
-        }
+    private static void checkJdksVersions(RootProject rootProject, int... versions) {
+        assertThatJdkDirectories(rootProject)
+                .as("JDK version directories match expected versions")
+                .containsExactJdks(versions);
 
         String osName = OperatingSystem.get().uiName();
         String archName = CurrentArch.get().uiName();
-        versions.stream().findFirst().ifPresent(version -> {
+        for (int version : versions) {
```
felixdesouza: I think this for this bit, you could have `JdkDirectory` have a method: `currentPlatformPath`.

If we made `JdkDirectoriesAssert` a `AbstractIterableAssert` or `AbstractListAssert`, you could end up with inlining `checkJdksVersions` to where it's used and the assertion will look something like this I think?

```java
assertThatJdkDirectories(rootProject)
  .as("JDK version directories match expected versions")
  .containsExactJdks(versions)
  .allSatisfy(jdk -> {
    assertThat(jdk.currentPlatformPath().resolve("download-url")).exists();
    assertThat(jdk.currentPlatformPath().resolve("local-path")).exists();
  });
```

felixdesouza: if you don't like the inference of `currentPlatformPath` we could just change it to:

```
platformPath(OperatingSystem os, CurrentArch arch)
```

felixdesouza: wdyt?

crogoz (dismissed):

gradle-jdks/src/test/java/com/palantir/gradle/jdks/GradleJdkPatcherIntegrationTest.java
```diff
@@ -105,7 +106,17 @@ void successfully_generates_gradle_jdk_setup_files(GradleInvoker gradle, RootPro
                     .as("gradlew has exactly one patch footer")
                     .containsOnlyOnce(GradleJdksPatchHelper.PATCH_FOOTER);
 
-            checkJdksVersions(rootProject, 11, 17, 21);
+            assertThatJdkDirectories(rootProject)
+                    .as("JDK version directories match expected versions")
+                    .containsExactJdks(11, 17, 21)
+                    .allSatisfy(jdk -> {
```
crogoz, line 112: this looks great! Thanks for putting this up.

crogoz: I wonder now, if we shouldn't actually check that we generated all the files for all `os-arch` pairs instead of only checking for the current architecture - but I can do this as a FLUP, not needed in this PR

PR #796, opened 2026-07-10
Do not persist workspace between check and publish in CI

arnav-ag:
## Before this PR

The persisting to workspace step was taking ~1.5x the time it took to actually perform the check. Example CI run: https://app.circleci.com/pipelines/github/palantir/gradle-jdks/4665/workflows/37b0f310-f865-4e9a-bc4f-c00ee65698bc/jobs/13815

Here we see the check step taking ~7.5 minutes whereas the persist step takes ~11.5 minutes. In this case, the persisted workspace wasn't even consumed afterwards, since this was simply an intermediate commit as part of a PR and not as part of a release.
## After this PR

Set `PERSIST_WORKSPACE_BETWEEN_CHECK_AND_PUBLISH` to `false`. In the worst case during publish, it will have to recreate the artifacts that were stored in the cache. The time taken to do a cold checkout + build during the publish step in CI is yet to be measured.
==COMMIT_MSG==
==COMMIT_MSG==

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,9 +63,6 @@ jobs:
             fi
       - run: ./gradlew --parallel --stacktrace --continue --max-workers=2 check -Porg.gradle.java.installations.fromEnv=JAVA_11_HOME,JAVA_17_HOME,JAVA_21_HOME,JAVA_HOME
       - *check-no-files-changed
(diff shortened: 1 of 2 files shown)


CRogers:
:+1:

